PR #1, opened 2017-04-05
As suggested I am adding a user story

musicnerd:
The website suggests for users to add their stories. There are a fairly large number of musicologists who are interested in how musical score elements align with other non-score elements. I've listed a few examples.


joeberkovitz:
It's a good case -- would you mind cross-linking this to RLP7 somehow, since I want to emphasize that there are multiple uses for this sort of linkage? (You can just use `<a href="#RLP7">...</a>`)

PR #71, opened 2018-02-20
Reference the HTML5 standard definition of semantic markup.

### LINKED ISSUE #50: What the W3C means by "semantic"
The definition of "semantic" in §2.7 of the new [Draft MNX Specification](https://w3c.github.io/mnx/specification/) is a bit vague:
> While it is challenging to ground the definition of what is "semantic" in music notation, the litmus test used here is the practical one of widely shared understanding: if the performance interpretation of some notation is widely shared within the culture targeted by some semantic module, then that aspect may be classified as semantic.

I think it would avoid confusion in future, if the word were used in the (rather specialised) sense used at the W3C. As I understand it, they are, after all, going to be the document's ultimate recipients. So I've been trying to find out what they mean by the word.

https://www.w3.org/RDF/Metalog/docs/sw-easy
begins:
> Many people don't have a clear idea of what the Semantic Web is.

Which confirms my opinion, that we cant define "semantic" simply in terms of a widely shared understanding. :-)

According to
https://en.wikipedia.org/wiki/Semantic_Web

The W3C says:
> The Semantic Web provides a common framework that allows data to be shared and reused across application, enterprise, and community boundaries

The article also says that:
> The term [Semantic Web]  was coined by Tim Berners-Lee for a web of data that can be processed by machines—that is, one in which much of the meaning is machine-readable.

and that Tim Berners-Lee has defined "The Semantic Web" as:
> A web of data that can be processed directly and indirectly by machines.
 
The whole Wikipedia article is well worth reading. I wouldn't be surprised to discover that it was largely written by Sir Tim himself.

So the "Semantic Web" is (or consists of) the W3C standards and, for the W3C, "semantic" means something like "A convention that has been agreed for the use of programmers when communicating between themselves and with machines."

That's all right if you are trying to describe the umbrella organisation but, as programmers, we don't need to refer to W3C standards _in general_.
For example: the _class_ attribute that can be defined on an SVG element is a _particular_ W3C standard. Saying that an object has `class="notehead"` says what the object _is_, but nothing about what it _means_ or _how it looks_. The class is just a hook that can be used to determine other attributes. So, while class information is "semantic" by definition (its a W3C standard), it doesn't help much to say so when discussing the use of a _particular_ class or any other _particular_ (potential) W3C standard.

For this CG's purposes, I'd suggest using the word in the following sense:
- "Semantic information is information that either conforms to a W3C standard, or can be made to do so in future."

or
- "Semantic information is information that has a precisely agreed meaning for both programmers and machines."

That's pretty close to what the current Draft MNX Specification says, but I think is important to include the reference to machines.

joeberkovitz:
Fixes #50

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/71.html" title="Last updated on Mar 15, 2018, 4:46 PM GMT (b97ac1c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/71/f9e38bf...b97ac1c.html" title="Last updated on Mar 15, 2018, 4:46 PM GMT (b97ac1c)">Diff</a>


notator:
In §2.7, under the heading "Separate semantic concerns from presentation/interpretation concerns.", it still says:

> While it is challenging to ground the definition of what is "semantic" in music notation, the litmus test used here is the practical one of widely shared understanding : if the performance interpretation of some notation is widely shared within the culture targeted by some semantic module, then that aspect may be classified as semantic.

This appears to be providing an alternative definition for "semantic". I think the paragraph should simply be deleted.

joeberkovitz:
I think that paragraph reiterates the HTML5 definition rather than conflicting with it, but for clarity it's been removed.

notator:
@joeberkovitz Thanks! :-)

PR #81, opened 2018-03-15
Generalize and sparsify global measure content

### LINKED ISSUE #2: Should global (system) notations be interspersed within regular parts?
Instead of placing within `<measure>` elements within `<global>`, should system notations be placed in individual parts and tagged in some way, perhaps via a `scope="global"` attribute or something similar?

This would avoid the problem of the `<global>` or similar element being very sparse with most measures containing nothing.

However, this approach could scatter system notations throughout the score in a happenstance way, since there is no particular part that makes sense for them to belong to.

While they could be forced to be placed in the first part, this leads to ambiguities as sometimes the first part is not visible on all systems. In MusicXML, the result is that system notations tend to appear in the topmost _visible_ part, which varies unpredictably from score to score.

joeberkovitz:
Fix #2.

Generalize the use of <global> for shared measure content that's not necessarily common to all parts.
Add the index attribute to <measure> to allow sparse encoding of measure content.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/81.html" title="Last updated on Mar 19, 2018, 4:10 PM GMT (a4e894f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/81/2d9cde4...a4e894f.html" title="Last updated on Mar 19, 2018, 4:10 PM GMT (a4e894f)">Diff</a>

### DIFF
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -1310,7 +1310,8 @@ tempo indications.
       <time signature="4/4"/>
     </directions>
   </measure>
-  ...Note the omission of measures 2 and 3, which would have no content...
+  <measure index="2"/>
+  <measure index="3"/>
   <measure index="4" barline="final"/>
 </global>
 ```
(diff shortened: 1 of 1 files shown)


joeberkovitz:
Duh... I've renamed the proposed `sequence` attribute to `index`, to avoid confusion with `<sequence>`. Also `index` is now optional since it makes for annoying clutter in simple, non-sparse scores if not actually needed.

Issue #98, opened 2018-04-14
Split MNX-Common and MNX-Generic into two separate formats


adrianholovaty:
I've been concerned about how confusing it would be for end users to have to know the difference between MNX-Common and MNX-Generic.

Faced with a "Export..." dialog box that has two separate MNX formats, the majority of non-technical users wouldn't have the slightest idea why to choose MNX-Common vs. MNX-Generic. Heck, I'm a programmer of music-notation software who has been following the specification's progress, and _even I_ would have a hard time explaining the difference to a layperson.

Some anecdotal evidence: For my own product, which ingests MusicXML, I've gotten anxious emails from customers asking "Should I export as compressed or uncompressed MusicXML?!" Fortunately, the answer there is "We support both" — because the underlying notation data is identical and the only difference is that one of the files is zipped — but this won't be the case for the two MNX formats, which are radically different in meaning and scope.

I chatted about this with @joeberkovitz a few days ago at Musikmesse, and he explained his assumption was that there would be a single container format, so end users would only need to "export as MNX." That makes sense at face value, and I suggest documenting that more obviously. But I think this may merely be kicking the can down the road.

I think it would be a great benefit to the project to have clear answers to the following questions:

- Are MNX-producing applications expected to export MNX files that contain both MNX-Common and MNX-Generic?
    - If yes, then that's likely a nonstarter... Apps that don't deal with graphical display will have no way to generate MNX-Generic, and apps that deal without semantics won't have a way to generate MNX-Common.
    - If no, then how does an MNX-producing application communicate the following to end users:
        - Which type(s) of MNX it's capable of producing.
        - The difference between the MNX flavors (so that users don't just blindly choose one).
- Given an MNX file, how will an end user be able to quickly determine whether it's MNX-Common, MNX-Generic or both?
- How does an MNX-consuming application communicate which flavor(s) of MNX it supports?

The more I consider this, the more I think MNX could be a colossal failure in the marketplace if we don't address these issues clearly and cohesively. :-/ I have come to believe that using a single user-facing "brand" to handle both MNX flavors would be a bad idea. The only way to avoid mass confusion would be to have clearly distinct concepts — distinct file extensions, distinct "Export as..." options, distinct "Import" options.

It could very well be that this is the plan (I hope it is!); if so, let's make that crystal clear in the documentation.

clnoel:
[...]

Issue #79, opened 2018-03-14
How to represent nonstandard alignments in CWMN


joeberkovitz:
Bach, Chopin, Couperin and many other composers created scores with sequences of note values that are aligned in ways that do not reflect their temporal positions if calculated in a standard way from the durations.

As at least one set of examples, Chopin frequently employed passages in which the last note of a triplet was visually aligned with the 16th of a dotted-8th/sixteenth pair.

CWMNX needs a set of approaches to be able to deal with both the visual and performance renditions of such scores.

The current suggested direction is to permit the temporal position of any event for alignment purposes to be overridden by a style property. Note that `<interpret>` already allows complete latitude with respect to performance interpretation.

joeberkovitz:
With regard to https://github.com/w3c/mnx/issues/74#issuecomment-372957167 and the ensuing discussion, I do not see anything in any of the compositional examples cited that suggests we need a different semantic structure for Baroque music or for earlier periods of proto-CWMN. What we need is a mechanism to break the default layout and performance-interpretation rules, when they need breaking.

clnoel:
To quote myself for further discussion, referencing this image posted by @notator:
![image](https://user-images.githubusercontent.com/6453511/37416643-539215a4-277c-11e8-856e-c81413e9b017.png)


> There are three voices (sequences). The RH-upstem voice has q-q.-e (3 beats), the RH-downstem voice has q.-e-q (3 beats), and the LH voice has e.-s-e.-s-e.-s (3 beats). Which you would think would be no issue, given the 3/4 time. However, the the RH-downstem 8th note is notated at the same time as the second LH sixteenth note, which is a sixteenth off of where it would be independently played. So do you play it at beat 1.5 as an 8th note (assuming that the notated vertical alignment was unintended), or beat 1.75 as a 16th (assuming that the notated q.-e was meant to be flexible)? There's a similar issue for the RH-upstem 8th note and the last LH sixteeth note. Once again, that becomes something that a musician, and therefore the encoder, would need to interpret. And something that later composers would probably make explicit, either by offsetting the 8th note to it's "proper" location, or by using a quarter note tied to a dotted-8th followed by a 16th. 

I agree @joeberkovitz that the alignment issues do not require a new variant of MNX, or even a new profile (the profile issues I was seeing had to do with time signature changes that weren't notated, and might change very frequently).

However, there are two cases an encoder might need for this case:
[...]

PR #147, opened 2019-04-10
Readme patch: MNX-Common encodes CWMN

notator:
I've closed the PR at ff6b3b0 in order to send PRs for the separate suggested changes. They should be easier to deal with individually. I'm also practising sending PRs! :-)
This is the first of the changes I suggested in ff6b3b0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Quick links:
 
 ## An introduction to MNX-Common
 
-MNX-Common is a new, open standard for representing Common Western Music Notation (CWMN) as machine-readable data. It’s still being created, but this document gives context on our goals and plans.
+MNX-Common is a new, open standard for representing music notation as machine-readable data. It’s still being created, but this document gives context on our goals and plans.
 
 Conceptually, MNX-Common is a way to represent this image:
 
@@ -29,7 +29,7 @@ The former, an image, is intended for humans. A computer can’t understand it w
 
 The latter — which encodes the _meaning_ (aka _semantics_) of the music, as opposed to a specific collection of pixels — is much easier for a computer to understand. Treating music in this way means it can be shared among different programs, automatically manipulated and much more.
 
-MNX-Common is a standard that abstractly encodes CWMN and its performance. CWMN notation editors will be able to create concrete score instantiations from MNX-Common, that will be consumable by web browsers and other software.
+MNX-Common is a standard that describes how to encode music, so that computers can have a shared understanding on how to interpret it. As HTML describes a web page such that any web browser can read it, MNX-Common describes a piece of music such that any notation software can read or write it.
 
 ### Motivation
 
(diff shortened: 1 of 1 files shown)


cecilios:
@notator 
I would suggest more meaningfull names for PRs, similar to those in issues. "Readme patch1" is quite obscure.

As to the proposal, I do not understand the need to narrow the definition of MNX-Common. In my oppinion, its objective should be to represent music notation as broadly as possible. But to be pragmatic, we agreed to narow MNX-Common applicability to specific 'music notations' by using profiles. Why to change current wording to include the acronim CWMN without defining what is included and not included? No gain with your proposal, as it doesn't define MNX-Common scope more than current wording, but it creates more doubts and limitations for future profiles, i.e. Is neumes or tablature included in your CWMN definition?

notator:
Hi @cecilios!
Okay, I've changed the title to be more expressive. Sorry, I'm new to PRs...

According to [the current draft spec](https://w3c.github.io/mnx/specification/#mnx-score-types), MNX-Common

> encodes Conventional Western Music Notation (CWMN) 

But the spec's [CWMN definition](https://w3c.github.io/mnx/specification/#cwmn) is far too vague to be used as the basis for a software standard. Last year, we had an intense, but inconclusive discussion (in [issue #79]( https://github.com/w3c/mnx/issues/79)) trying to pin CWMN down... 

I outlined a solution to this dilemma  a couple of weeks ago [here]( https://github.com/w3c/mnx/issues/98#issuecomment-476537369).
Basically, I think the problem can be resolved by defining CWMN as the notation implemented by MusicXML, Finale, Dorico, Capella etc. If in doubt, the manufacturers should tell us what their definition is. This seems to be the best practical way forward, In practise, MNX-Common is currently being defined as the successor to MusicXML, without reference to any of the world's other notations, and that's fine by me...

The spec is also rather vague about the difference between a "notational idiom", a "score type" and a "profile". This also needs clearing up, and is also covered by [my last reply to issue #98](https://github.com/w3c/mnx/issues/98#issuecomment-476537369). Lets continue the discussion there.

cecilios:
Hi @notator,
One thing is the specification, in which all kind of details and explanations can be included, and a different thing is the Readme.md page, that will be, in many cases, the first page someone approaching MNX will see. The spec. is the best place for details about MNX scope.

We should not start the Readme.md page talking about CWMN in the first paragraph without having previously defined the precise scope of the group understanding of what is CWMN. It is true that a few paragraphs later, the Readme.md text introduces CWMN and explains more about this. But it doesn't limit MNX to CWMN, it says:

`Our goal with MNX-Common is to create a format that does all of the following: ...
- It supports all of Common Western Musical Notation (CWMN)
...`

That is, the goal is to **at least** allow for CWMN. It doesn't not say **to limit MNX to CWMN**.

In any case, even assuming that the objective could be to limit MNX to CWMN,  I would not start MNX introduction with acronyms and scope limitations before the reader has got a better understanding of MNX purpose and objectives. The readme.md page So, IMO we should not enter into details in the Readme.md page; the right place for details about MNX scope  profiles, etc. is the spec.

notator:
I agree completely. The Readme could/should really be a much more general introduction.

It could be very like the text at the top of [the W3C page](https://www.w3.org/community/music-notation/). However, that page also needs updating, since it contains a link to this MNX repository but doesn't say what the repository contains. In particular, the text should make it very clear (especially for the W3C audience) that MNX **is not limited to CWMN**. It should definitely **_not_** say that

> MNX-Common is a new, open standard for representing music notation...

We should really continue the discussion about the spec in Issues #98 and #138. These are currently volatile and under Active Review. But here's a bit of background that might help understand how the new Readme came to be the way it is:

At the face-to-face meeting in Frankfurt (see [YouTube video](https://www.youtube.com/watch?v=QA86VOrN-u0)) @mdgood proposed that MusicXML should continue to be supported with maintainance updates, but that MNX-Common should be a new standard that takes more modern practices into account, and encapsulates any new ideas coming from the MusicXML community. The hope is that MNX-Common will eventually be able to replace MusicXML altogether. There was general agreement to this proposal, as you can see in the video.

But MNX is also supposed to be about notations that are not covered by MusicXML. This is, after all, a W3C Community Group! I'm not sure if MusicXML covers Gregorian Chant, other forms of neumes or piano-roll notations, but I'm pretty sure it does _not_ cover Asian notations that read vertically... Such notations need to be handled in a detailed, semantic way, just as much as CWMN, so MNX-Generic is not the answer there. (@adrianholovaty  is currently separating MNX-Generic into a separate Draft Specification document.)

The co-chair has said that they want us to concentrate on making progress with MNX-Common, so that's what we are currently doing. While we are doing that, I'd like to ensure that the (MNX) context within which MNX-Common is defined will easily accommodate non-MNX-Common notations when the time comes. That means defining generally useful concepts such as MIDI.cent frequency notation  ([Issue #138](https://github.com/w3c/mnx/issues/138#issuecomment-480789811)), and clearing up the muddle surrounding _score types_, _notation idioms_ and _profiles_.

Apropos profiles: In [Issue #138](https://github.com/w3c/mnx/issues/138#issuecomment-480789811), we are, for example, currently discussing the MNX-Common syntax for microtone symbols. This has to be done, whatever happens, but its quite possible that such symbols don't belong in all the MNX-Common profiles. What each MNX-Common profile actually contains (and how many profiles there are going to be) can be decided later.

PR #151, opened 2019-04-27
New ReadMe

notator:
This is a completely new, much simplified ReadMe, that describes more clearly where we are really at.
It is similar to the ReadMe files in the [MusicXML](https://github.com/w3c/musicxml) and [SMuFL](https://github.com/w3c/smufl) repositories.

Comments and suggested further improvements welcome!

This pull request overrides https://github.com/w3c/mnx/pull/147, but I've left that open because the comments there are still relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,66 +1,15 @@
 # MNX
 
-Welcome to the [Music Notation Community Group](https://www.w3.org/community/music-notation/)'s next-generation music markup proposal.
+Welcome to MNX, the [W3C Music Notation Community Group](https://www.w3.org/community/music-notation/)'s next-generation music markup proposal.
 
-Quick links:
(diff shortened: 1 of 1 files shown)


dspreadbury:
I don't like this change. @adrianholovaty worked hard on the last revision of the ReadMe and the co-chairs agree that it serves a useful purpose. I would oppose accepting this pull request.

adrianholovaty:
Not only do I not see the need for this change, I also think the specific text changes are markedly worse. It's harder to understand, removes nearly all the context and uses unnecessarily technical/overwritten language.

The goal of the new Readme was to provide some context on why we're designing a new notation format, written in an approachable way. This pull request destroys that.

@notator, you appear to be trying to solve a non-existing problem here. I worry that you're wasting our time.

notator:
This rejected ReadMe can still be read at
https://github.com/notator/mnx/tree/patch-1

Issue #19, opened 2017-04-28
Allow pitch encodings to employ cents or decimal semitones?


joeberkovitz:
From @mogenslundholm:

Would it [be possible to] write: <note pitch="C4+4,98044999134613"/> ?
The tone <note pitch="F4-0,01955000865387"/> will give same result.
Refer to "MNX Proposal Overview": 5.3.10. Pitch encoding

(The alter values written by plus- or minus-sign are in half steps, i.e. 12 halftones per octave. And 12*log2(4/3)=4,98044999134613).

joeberkovitz:
This is possible. It's also possible we could use cents, since many temperaments have been analyzed in terms of that unit (I realize it's just a matter of where the decimal point goes)

cecilios:
An equally tempered semitone is equal to 100 cents. Therefore, as you noted, using cents or decimal semitones is just displacing the decimal point.

Adding something (i.e. a character at the end) to encode the type of units (cents or semitones) would complicate exporters and parsers with no real gain. So I would vote for allowing only one type of units in the standard, either cents or semitones.

natiwipi:
I totatally agree with you, cecilios.

clnoel:
I am commenting here to bring this back up to the top of the recently-commented list, since some of the discussion in #138 has to do with how to do pitch representation, both in pitch-spelling and (as this seems to be) sounding-pitch. Finding a better way to represent microtones than the four different methods [in the current spec](https://w3c.github.io/mnx/specification/#chromatic-pitch-syntax) seems important to me!

Using MIDI-cent (60.5) or SPN-cent(C4.5) notation seems to be up there.

--Christina

notator:
@cecilios said
> I would vote for allowing only one type of units in the standard, either cents or semitones.

Agreed.
In #138, we are currently assuming that `<note>` is going to have a separate (optional) `sounding` attribute that would determine its frequency. The frequency information can't be located in the (graphic) `pitch` information since it depends on external state information (the parser's current key, transposition and measure state).  The `<note>`'s `sounding` attribute, if it exists, would _override_ the frequency calculated using the `pitch` attribute and the current state of the parser.
This simplifies things considerably, since the `pitch` attribute no longer has to contain both graphical and temporal information (as in [this issue's opening comment](https://github.com/w3c/mnx/issues/19#issue-225078473)). 

@clnoel raises the question of whether we should use MIDI-cent (60.5) or SPN-cent(C4.5) notation.
[...]

Issue #4, opened 2017-04-25
Are notes encoded in sounding or written pitch?


joeberkovitz:
In MusicXML, all pitches are framed in terms of the instrument's transposed pitch. In MNX, we can re-examine this question.

Encoding in transposed pitches makes performance interpretation into more work since the instrument's transposition must be applied, and this transposition could even vary during the course of a piece.

It also makes dynamic display of concert pitch trickier, since assumptions must be made about enharmonic spelling and key signatures.

On the other hand, encoding transposed pitches will more accurately reflect a manuscript that has been composed in transposed form, and may ultimately be more robust in terms of delivering a definitive final instrumental part.

Other pros and cons need to be brought out also.

mogenslundholm:
I feel I am not sure about the meaning of the words: Let say we have a pitch, that we measure by an oscilloscope or tuner when the note is played on the instrument, the real pitch - is this "in concert".? Anyway, I think the real pitch value should appear in MNX. 
As I remember it - MusicXML is different, but there is also the symbol "Ottava" or "8va" and in this case MusicXML uses the real pitch values.  Also different Clef-symbols will have same pitch-values, but the notes will look different on the paper.

siennamw:
Perhaps this topic should be renamed, "Are pitches encoded in sounding or written pitch?" for clarity.

I can see advantages and disadvantages with both approaches.  Perhaps both should be supported.  Each part could be flagged as being encoded with either sounding or written pitches, with `<part pitch-encoding='sounding'>` or something similar.

joeberkovitz:
I've renamed the topic for greater clarity. Noted that octave displacements play a role here as well.

@mdgood, can you chime in at some point with an explanation of the design rationales for MusicXML's pitch encoding so we have your background on this?

mogenslundholm:
Two ways to write transposed pitch? No, I would not like that. This would double testing and increase risk of errors, because properly one of them will be used seldom and therefore not tested in practice. With only one way to do it, this way will be tested every time.
(For me the sound is the real pitch - but I can add an offset like with MusicXML)

siennamw:
I'm still not sure that we're talking about the same thing here yet @mogenslundholm.  The question is, how do we encode music for transposing instruments?  When we write for a clarinet in B-flat, for example, if we want to hear a B-flat are we encoding B-flat (sounding pitch) or C (written pitch)?

mogenslundholm:
[...]

PR #152, opened 2019-05-01
Clarify whether 8va affects definition of 'written pitch'

adrianholovaty:
This pull request ties up a loose end from commit cf46aa48a2f1ce21df5f4f4cac4d37b6f3757ea5 — which defined "written pitch" but didn't specify how/whether ottava markings affect the written pitch from MNX-Common's perspective.

This change to the spec uses the same approach as MusicXML, which is that ottava markings are considered presentational and don't affect the "written pitch." Here's some relevant thinking from Michael Good (via email), musing on MusicXML's experience:

> In MusicXML, the <pitch> element represents is the semantic concept of pitch for someone reading the music, not a graphic representation of where the pitch happens to be notated on the staff. There is no transformation required for accidentals present in key signatures (those alterations have to be included in the pitch even if no accidental is displayed), octave lines, or octave-transposing clefs. The only transformation needed comes from data in the <transpose> element.
>
>This part of MusicXML has seemed to work really well, so I believe it should be maintained for MNX-Common.

I'd like to get two bits of feedback on this:

1. Is there any opposition to the treatment of ottavas as presentation (i.e., not affecting written pitch)?
2. Is the specific wording clear and unambiguous enough?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/152.html" title="Last updated on May 2, 2019, 9:43 AM UTC (c1dacbf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/152/cf46aa4...c1dacbf.html" title="Last updated on May 2, 2019, 9:43 AM UTC (c1dacbf)">Diff</a>

### DIFF
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -979,9 +979,12 @@ the pitch that an instrument generates.
 A note's written pitch differs from its sounded pitch in case of a
 part written for a transposing instrument, such as a clarinet.
 
-<em>Needs clarity about how 8va markings are taken into account. Is the
-written pitch the literal pitch that's printed on the page, or is it the
-pitch that the musician is "thinking of" (after applying the 8va)?</em>
+MNX considers ottava markings (e.g., "8va," "8vb") to be purely
+presentational. Hence, ottava markings have no effect on a note's
+written pitch. For example, if a note is rendered on the bottom
+staff line without an ottava marking, and a second note is
+rendered on the top staff space with an "8vb" marking, the two
+notes have the same written pitch.
 
 <h4 id="common-notational-syntaxes">Notational syntaxes</h4>
 
--- a/specification/index.bs
+++ b/specification/index.bs
@@ -972,12 +972,15 @@ and articulations accordingly.</li>
 
 <h5 id="written-pitch">Written pitch</h5>
 
-A note's <dfn>written pitch</dfn> is the pitch that is rendered in
-notation. This might be different from its sounded pitch, which is
-the pitch that an instrument generates.
+A note's <dfn>written pitch</dfn> is the pitch that would sound
+if the note's notation were performed by a concert-pitch instrument.
+Written pitch can be thought of as the note's pitch from the
+musician's instrument-specific perspective.
 
-A note's written pitch differs from its sounded pitch in case of a
-part written for a transposing instrument, such as a clarinet.
+A note's written pitch might be different from its sounded pitch,
+which is the pitch that an instrument generates. Written pitch
+differs from sounded pitch in case of notation written for a
+transposing instrument, such as a clarinet.
 
 MNX considers ottava markings (e.g., "8va," "8vb") to be purely
 presentational. Hence, ottava markings have no effect on a note's


notator:
Feeedback 1:  The answer is yes. We are currently still debating this change in #4, So this PR should not be applied until we have reached agreement there.

Feedback 2: The spec would read as follows after applying this PR:

>**§ 5.2.1.12. Written pitch**
>A note’s written pitch is the pitch that is rendered in notation. This might be different from its sounded pitch, which is the pitch that an instrument generates.
>
>A note’s written pitch differs from its sounded pitch in case of a part written for a transposing instrument, such as a clarinet.
>
>MNX considers ottava markings (e.g., "8va," "8vb") to be purely presentational. Hence, ottava markings have no effect on a note's written pitch. For example, if a note is rendered on the bottom staff line without an ottava marking, and a second note is rendered on the top staff space with an "8vb" marking, the two notes have the same written pitch.

The first sentence defines "written pitch" as the pitch that is rendered in notation (i.e. in the printed score), but the "written pitch" in the final paragraph (this PR) seems to be referring to the `pitch` attribute of a `<note>` in the XML. That's confusing, because the XML is not the notation.

If you really mean the `pitch` attribute here, then I think this clarification belongs in the definition of that attribute (in [§5.2.6.1](https://w3c.github.io/mnx/specification/#the-note-element)). You need to be careful not to define "written pitch" in terms of the `pitch` attribute, and then define the `pitch` attribute in terms of "written pitch".

I'd recommend waiting for the outcome of #4 before spending time on addressing these objections. :-)

adrianholovaty:
@notator: Thanks for the feedback. I believe we already have group consensus on #4 — that is, we will be treating the written pitch as the primary/required pitch for each note, rather than the sounded pitch (with the specific encoding of the latter being yet undetermined).

Can you outline your specific opposition to the treatment of ottavas as presentation?

That's a good point about the ambiguity of the text. I've just taken another stab at it and updated this pull request:

> A note's written pitch is the pitch that would sound if the note's notation were performed by a concert-pitch instrument. Written pitch can be thought of as the note's pitch from the musician's instrument-specific perspective.
> 
> A note's written pitch might be different from its sounded pitch, which is the pitch that an instrument generates. Written pitch differs from sounded pitch in case of notation written for a transposing instrument, such as a clarinet.
> 
> MNX considers ottava markings (e.g., "8va," "8vb") to be purely presentational. Hence, ottava markings have no effect on a note's written pitch. For example, if a note is rendered on the bottom staff line without an ottava marking, and a second note is rendered on the top staff space with an "8vb" marking, the two notes have the same written pitch.

Feedback from all welcome!

notator:
@adrianholovaty 
> Can you outline your specific opposition to the treatment of ottavas as presentation?

First: I'm not alone in this. I'm supporting the proposals @clnoel and I worked out in #138 (beginning [here](https://github.com/w3c/mnx/issues/138#issuecomment-480789811)).  

Basically, and rather roughly, I think that our proposals work better than MusicXML's approach, especially when it comes to transposition and the representation of microtones (accidentals etc.). The two approaches are almost symmetrical, but only almost.

As @clnoel said in [Issue #138](https://github.com/w3c/mnx/issues/138#issuecomment-480876552)
> I support the idea of a transposition as a direction. That makes a LOT of sense.

If the best way to handle transposition is for the parser to track the values of a `<transposition>` `<direction>`, then it would be more consistent, and intuitively easy to understand, if the values of _all_ the relevant `<direction>` elements (`<ottava>`, `<key-signature>` (etc.?)) were accumulated and used by the parser in the same way. This is the way humans parse scores, so its very easy to understand.

I have a refinement to add, to @clnoel's and my proposals, about non-standard accidentals, but I think I should keep the proposals all in one place and do that in #4. 

In other words, while (as @mdgood says) MusicXML's approach may have been working well, I don't think its the best possible solution going forward.

You said
> A note's written pitch is the pitch that would sound...

I think this confuses space and time. A written pitch can't be a sounding pitch! A written pitch is something that is on paper in a score, or writing on a screen in an XML file, having spatial units of measurement (pixels, centimetres). The pitch that sounds has temporal units (Hertz etc.). So you can't define one in terms of the other. A meter can't be defined in terms of seconds. Its a "philosophical problem"! :-)

dspreadbury:
The real philosophical problem is the assumption made by @notator that everything can be split into notions of space and time. I find this highly problematic and I doubt I am alone. @notator, there is no general agreement that the primary concern of our encoding methods should be to separate the concepts of space and time. I don't think it's helpful to say that a written pitch exists in space and a sounding pitch exists in time, even if it may, if you look at it a certain way, be true.

Can you please try to use terminology that is easier for others to agree with and which doesn't make this broad binary division that forces everything to be one thing or the other? Thanks for your consideration.

notator:
@dspreadbury The problem with ignoring the difference between space and time in music notation/performance, is that one ends up going round in circles. Being careful about such things is a great way to make progress. Especially in a context like this, where we are trying to develop a formal specification document. If there are ambiguities in the root definitions, we will have endless trouble agreeing on what we are talking about, and people will eventually get bored and drift away. If there are contradictions, the whole spec will eventually collapse. So its important not to gloss over the basic definitions.

That having been said, I completely take your point about not unnecessarily complicating the discussion, and trying to use terminology that is easier for others to agree with. I'll try to do better in future.

Suggestion: Perhaps "written pitch" could, from the beginning, be defined in terms of the [Scientific pitch notation](https://en.wikipedia.org/wiki/Scientific_pitch_notation) standard, and its relation to frequency?

clnoel:
I have responded into #4. There is an underlying assumption issue we need to resolve before we can commit this!

notator:
In [my first comment above](https://github.com/w3c/mnx/pull/152#issuecomment-488596781), I said:

> If you really mean the pitch attribute here, then I think this clarification belongs in the definition of that attribute (in §5.2.6.1).

@adrianholovaty 
I think you could avoid having to define "written pitch" altogether. You could, delete [§5.2.1.12. Written pitch](https://w3c.github.io/mnx/specification/#written-pitch) completely, and confine the proposed clarification to the definition of 'pitch' in [§5.2.6.1 The `<note>` element](https://w3c.github.io/mnx/specification/#the-note-element). You could use a reference  to [Scientific Pitch Notation](https://en.wikipedia.org/wiki/Scientific_pitch_notation) instead of using the undefined (and confusing) "written pitch".

adrianholovaty:
@notator The goal is to disambiguate between written and sounding pitch, so that developers using MNX-Common can unambiguously know which of the two pitches to encode. The spec already references Scientific Pitch Notation in the [chromatic pitch section](https://w3c.github.io/mnx/specification/#chromatic-pitch).

notator:
@adrianholovaty I've just searched the current spec for "written pitch". The only places it occurs are in its definition in §5.2.1.12 and in the definition of 'pitch' in §5.2.6.1.
Since the Scientific Pitch Notation link is being used elsewhere in the spec, you could just reduce the number of entities you are defining (Occam's Law)...

adrianholovaty:
@notator That's a fair point, but I'm inclined to keep them separate and tightly defined concepts rather than overloading "pitch" — in case future parts of the spec need to refer to one concept but not the other.

notator:
@adrianholovaty Fine! :-)

shoogle:
@adrianholovaty:

__Feedback 1:__ I agree that ottavas should be treated as presentational rather than semantic.

__Feedback 2:__ I think the proposed wording is unnecessarily complicated. I would restore the first and second paragraphs to how they were originally and just keep your explanation of how ottava markings are handled.

I agree that the term "written pitch" is ambiguous. I [suggested](https://github.com/w3c/mnx/issues/4#issuecomment-487327533) using the term "transposed pitch" instead, but I also like [your suggestion](https://github.com/w3c/mnx/issues/4#issuecomment-489050546) of "Performer-Centric Written Pitch". The original spec refers to this as:

> the pitch that the musician is "thinking of"

Perhaps this could be called "conceptual pitch"? It is the pitch name that musicians would use to refer to the note, which may be different to how it sounds or is displayed.

BTW, it may be worthwhile to link back to relevant issues from within the specification itself. E.g:

> The pitch attribute MUST contain a transposed pitch (see Issue #4).

And

> We avoid using the terms __written pitch__ or __sounding pitch__ as these are ambiguous (see PR #152).

This provides a way for people reading the specification to find out why particular decisions were made without having to clog the specification with lengthy explanations.

notator:
> Perhaps [the pitch that the musician is "thinking of"] could be called "conceptual pitch"? It is the pitch name that musicians would use to refer to the note, which may be different to how it sounds or is displayed.

(Please don't bite my head off!)
Musicians usually refer to pitches with phrases such as "The C above middle C".
Seriously: I don't see how a pitch name that is "different to how it sounds or displayed" can be relevant to an XML definition. So I don't think it belongs in the spec. The spec becomes much simpler without it, 

Can we agree that **pitch names** are defined by Scientific pitch notation (e.g. "A4" is the **name** of the pitch (frequency)  440Hz)?

notator:
The ambiguity in the proposed "written pitch" definition is that it confuses the written score with the written XML. I think this can be resolved as follows:

XML is text, so the "written pitch" there should be a string (e.g. "A4").

The "written pitch" in a score is a graphic that depends on the clef. See https://github.com/w3c/mnx/issues/4#issuecomment-489331528

So the two are very different. I'm not sure if/how that insight should be put in the spec.
My inclination would be to keep any discussion about the written **score** out of the XML definition, and to make it very clear that "written pitch" refers only to the XML. In the current spec, "written pitch" is only used in the XML `pitch` definition.

notator:
With regard to the current PR, I think the answer is that the way 8vas work does _not_ affect the definition of "written pitch", so you could simply delete the final paragraph of §5.2.1.12.
The way 8vas work is a different issue, currently being discussed in #4.

dspreadbury:
@notator wrote:

> The "written pitch" in a score is a graphic that depends on the clef. See [#4 (comment)](https://github.com/w3c/mnx/issues/4#issuecomment-489331528)

No, the pitch does not depend on the clef. C4 would always be middle C regardless of the clef used.

notator:
The _frequency_ does not depend on the clef.
> C4 would always be middle C regardless of the clef used.

Agreed. But
> The "written pitch" in a score is a graphic that depends on the clef. See https://github.com/w3c/mnx/issues/4#issuecomment-489331528

A C4 notated in the bass clef is one ledgerline _above_ the staff. A C4 notated in the treble clef is one ledgerline _below_ the staff. The vertical position of the notehead in the written score depends on the clef.

dspreadbury:
@notator wrote:

> The vertical position of the notehead in the written score depends on the clef.

That's right: the vertical position of the notehead depends on the clef. But this is not its "written pitch", of course.

notator:
Can you say what you think the "written pitch" is?

shoogle:
The term "written pitch" is defined in the specification as being the pitch name (e.g. "C4"). It's true that you could encode it another way that is closer to how would be displayed in a printed score (e.g. "1st ledger line below the staff"). This might even make more sense in some situations (e.g. OMR) but not for composition or general useage because you would have to rewrite all the pitches every time you decide to change the clef. "C4" is semantic whereas "1st ledger line below the staff" is presentational.

notator:
@shoogle 
> The term "written pitch" is defined in the specification as being the pitch name (e.g. "C4").

That's definition we need to clarify.

Maybe we could agree that "C4" is meant, both when we read "C4" in the XML and when we read "treble clef, one ledgerline down" in the score.

The confusing thing about the current definition of "written pitch" is that it immediately raises, and does not answer, the question "Written where?". However, the spec is actually concerned only with defining an XML, so we don't really need to concern ourselves with the score representation here. Which means, I think, that §5.2.1.12 should simply define the different ways in which pitch is written in the spec (XML).

Can we agree that [Scientific pitch notation](https://en.wikipedia.org/wiki/Scientific_pitch_notation) defines **pitch names** and their associated **frequencies**? For example, "A4" is the **name** of the frequency 440Hz.

The way to represent a **frequency** in the spec (XML) is currently being discussed in #19. I think the result of that discussion also belongs in §5.2.1.12.

---

About this PR's title:
https://github.com/w3c/mnx/issues/4#issuecomment-489331528 describes why, in the proposal I'm defending, I think 8vas should have no effect on the representation of pitch in the `<note>`  element. That's because the two parsers, parse the file in the same way as a human would (i.e. the frequency of a C4 notehead is transposed up by an octave if there is an 8va in force.)
That proposal currently describes how to deal with several things that are not covered by the proposal in the spec: 8vas, transposition, precise tunings (microtones), non-standard accidentals. The spec's proposal has some catching up to do.

clnoel:
@shoogle Since we are in the process of defining written and sounding pitches in the current spec, using the existing definition of "written pitch" in the spec as the definition of written pitch in this thread is circular.

That being said, I get your point: You define written pitch as the pitch a performer would claim (clef+key+accidental+octava+space/line). And you are recommending recording this as letternote+accidental+octave (e.g. 'Eb5') in the XML, which is a slightly different issue.

I don't think everyone agrees that this _is_ the written pitch. Technically, the "written pitch" is an image on a page, which is _way_ too far towards the graphical side of things. We have already decided that performed frequency (that takes everything into account all the way through transposing instruments and unwritten microtones) is _way_ too far toward the audible side of things, and put that in a separate optional property (sounding pitch). See #4 for more comments in this line.

shoogle:
> using the existing definition of "written pitch" in the spec as the definition of written pitch in this thread is circular.

True, but I wasn't trying to define it, I was merely pointing out that it already had a definition. My comment was directed at @notator who had [asked](https://github.com/w3c/mnx/pull/152#issuecomment-489713505) @dspreadbury to clarify what he meant by "written pitch", when it was obvious that he was going by the definition in the spec.

> You define written pitch as the pitch a performer would claim (clef+key+accidental+octava+space/line)

No, that is how the spec defines it. I agree with you that "written pitch" is an image on a page, hence I [wanted](https://github.com/w3c/mnx/pull/152#issuecomment-489491791) to call the performer's kind "conceptual pitch" rather than "written pitch". Maybe "abstract pitch" would be even better?

> you are recommending recording this as letternote+accidental+octave (e.g. 'Eb5') in the XML

I am indeed, and this matters more to me that whether we decide to call it "written", "conceptual" or "abstract" pitch.

notator:
@shoogle 
I think it might be easier for a parser (any parser) to scan "Eb5" as separate entities (as MEI does) like this:
```
<note pitchClass="E" octave="5" accidental="b" ... />
```
Would you agree to that?

shoogle:
@notator, it would indeed be easier, but only marginally and not enough to justify the added visual complexity in my opinion. However that is off-topic for this thread.

Issue #111, opened 2018-05-11
Define how to encode instrument transposition


joeberkovitz:
This issue notes the need to represent the transposition of instruments including both key changes and octave shifts.

This is related to the more specialized concerns in #4 and #34 but concerns the general nature of the transposition.

clnoel:
I am commenting here to draw attention to the fact that there is an issue already to deal with these specifics, since part of the proposals in the general issue #138 include the addition of a `<transposition>` open-ended direction, which can handle several of the cases here. We should probably discuss here how to use that in conjunction with visible octava, key-changes, or clef-changes.

adrianholovaty:
See pull request #158 for some work on octave shifts.

adrianholovaty:
(Just renamed this issue to make the intent clearer.)

adrianholovaty:
Marking this closed in favor of #287.

PR #158, opened 2019-08-27
Add 'Octave shifts' to spec and by-example page

adrianholovaty:
In this pull request, I've added an "Octave shifts" section to the spec and the MNX-Common By Example page.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/158.html" title="Last updated on Sep 24, 2019, 9:20 AM UTC (38cca41)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/158/71ac02d...38cca41.html" title="Last updated on Sep 24, 2019, 9:20 AM UTC (38cca41)">Diff</a>


adrianholovaty:
This relates to issue #111.

adrianholovaty:
Here's a way to view the "MNX-Common by example" page as it stands on this branch:

https://cdn.githubraw.com/w3c/mnx/octave-lines/by-example/index.html

notator:
The link to "MNX-Common’s _definition of written pitch_" in the final paragraph of the _Octave shifts_ example in [MNX-Common by example](https://cdn.githubraw.com/w3c/mnx/octave-lines/by-example/index.html) is broken.

There are two problems with:
>notes are rendered down an octave from their true pitched values
1. The definition is incorrect if the 8va is being applied to a transposing instrument.
2. "true pitched value" has not been defined, so this could easily be a dimension mismatch (you can't render anything (graphically, in space) an octave lower than a frequency. So, if the rendering is graphical, then "true pitched value" needs to be explicitly defined in terms of graphics. Definitions should always use terms that are unambiguous.

Actually, I think the spec should contain a hierarchically organised _Definitions_ section, in which the terms "written pitch" and "sounding pitch" are defined as fundamental terms that can be used in dependent definitions.
The current [§4.12 written pitches](https://w3c.github.io/mnx/specification/common/#written-pitches) is rather informal and very confusing (I don't understand the first two paragraphs). The definition begins by saying that written pitch is a frequency:
>A note’s written pitch is the pitch that would sound if the note’s notation were performed by a concert-pitch instrument.

I think this "written pitch" must be the pitch written in the XML, not the pitch written on paper, but I'm not sure....

adrianholovaty:
> The link to "MNX-Common’s _definition of written pitch_" in the final paragraph of the _Octave shifts_ example in [MNX-Common by example](https://cdn.githubraw.com/w3c/mnx/octave-lines/by-example/index.html) is broken.

Yes — all the relative links in that page are broken when viewed on the githubraw.com domain. To get the correct link, replace cdn.githubraw.com with w3c.github.io. Here's the correct link: https://w3c.github.io/mnx/specification/common/#written-pitches

Regarding "notes are rendered down an octave from their true pitched values," I took that wording from [the MusicXML docs here](https://usermanuals.musicxml.com/MusicXML/Content/EL-MusicXML-octave-shift.htm). This is surprisingly tricky to define in a precise way. Do you have a concrete suggestion on a specific wording?

notator:
On second thoughts, maybe we only need a _Basic Definitions_ section...

Sorry this posting has got a bit long, but it ends with the answer to your question, -- and you did ask... :-)

The following definitions need to be discussed. There will be changes and refinements to be made, but I think the general idea is okay.

Definition: **pitch name**: 
A **pitch name** is a string consisting of:
an upper case letter in the range [A..G],
optionally followed by one of the following strings [#,b,##,bb ... ] denoting an accidental*
followed by a numeral character in the range [0..10] denoting the octave**

Examples of **pitch names** are A3, Gb5, F#3, Ebb3 etc.

Comment 1:  This definition ought to be formalised as a [regular expression](https://en.wikipedia.org/wiki/Regular_expression). I've done it less formally for the moment.
Comment 2: This definition is related to [§5.4. Chromatic Pitch Syntax](https://w3c.github.io/mnx/specification/common/#chromatic-pitch-syntax). That section also adds a specific frequency alteration after the octave number. Adding a specific frequency alteration to **pitch name** would create a type mismatch in the definition of **written pitch** below.
*Comment 3: I think the string denoting the accidental should be defined for all the accidentals in the SMuFL standard... (discuss)
**Comment 4: [Scientific Pitch Notation](https://en.wikipedia.org/wiki/Scientific_pitch_notation) defines 12 octaves (-1..10). Do we need -1? (MIDI 0 is C-1, which might be a parsing problem.)


Definition: **written pitch**
A **written pitch** is a **pitch name** associated with a graphic in a score (space).

Examples of **written pitch** are identical to the **pitch name** examples.
Comment: Each **written pitch** can be associated with many different graphic configurations. The following (from https://github.com/w3c/mnx/issues/4#issuecomment-489331528 ) are all representations of C4 (middle C):
image
An important omission from the above list is a base clef staff with a notehead in the third space down having an 8va above it. Such a diagram also represents the **written pitch** C4. We can even say, for short, that all such diagrams represent _written C4_ without being misunderstood.

Definition: **sounding pitch**
A **sounding pitch** is a **pitch name** associated with a _frequency_ (time).

Examples of **sounding pitch** are identical to the **pitch name** (and **written pitch**) examples.
Comment: Each **sounding pitch** can be associated with many different frequencies depending on the current tuning scheme and/or context (choirs, string quartets). [Scientific Pitch Notation](https://en.wikipedia.org/wiki/Scientific_pitch_notation) defines the frequencies associated with **sounding pitch** in standard 12-tone equal temperament tuning.
Note that all frequencies can always be named (using the currently available **pitch names**). In a (jazz?) performance of a score that uses standard accidentals and equal temperament by default, any frequency between about 427Hz and 453Hz would be classed as a (more or less flat or sharp) A4.

If the **written pitch** is supposed to be equal to the **sounding pitch** when a score is performed, then we can say that the score "sounds as written", i.e. is "non-transposing". The alternative is that the score is "transposed", and there will be a constant interval (frequency relationship) between the frequency associated with the **written pitch** (interpreted as a **sounding pitch**) and the frequency associated with the **sounding pitch**.

--
@adrianholovaty 
> 8va: notes are rendered down an octave from their true pitched values

Can now be rephrased as:
"8va: raises both **written pitch** and **sounding pitch** by one octave.
or
"8va: both **written pitch** and **sounding pitch** are one octave higher than if the 8va were absent."

Comment 1: To raise any **pitch name** by one octave, just increment its octave index by one.
Comment 2: Note this also works for transposing scores (in which the **written pitch** and **sounding pitch** are different).

--

Question: Are the signs really correct in the pull request? Currently "-8" means 8va, "8" means 8vabassa, "-15" means 15ma and "15" means 15mabassa? That's rather counterintuitive. Maybe these are typos?

mscuthbert:
Just FYI -- in music21 we have a default setting ("graphical") but allow "graphical" or "transforming" (or some terminology like that) as settings.  Graphical ottavas expect the pitch of the note element to be the sounding pitch and thus change the way the note looks.  Transforming ottavas leave the note in the same place on the staff but change how the note sounds.

The reason for having both is to encode ambiguous situations where it is unclear whether a note should be covered by the ottava or not, where it seems more truthful to encode the notes at written pitch and let the user manipulate the ottava spanner in order to determine which notes are covered by the ottava or not.

Another example where transforming ottavas are helpful: imagine a short score / particel which encodes the melody but puts in "tpt +fl 8va"  -- this would be a specialized ottavation that should not need the notes to be encoded twice an octave apart.  (this would not be covered by this spec, but something to think about in the future).

I could not find any documentation about what MEI expects for what they call "octavation".

mscuthbert:
Btw -- is there a text attribute or anything that specifies how the octave-shift is to be displayed?  A number of scores use "15va" instead of "15ma" or "8va bassa" instead of "8vb" etc.  Cage uses "16va" (or 16ma, I can't remember) instead of 15va.  etc.

btw I think this is incorrect:  "8va: raises both written pitch and sounding pitch by one octave."   One of the two must remain the same under 8va.  8va either LOWERS written pitch by an octave ("graphical") or raises SOUNDING pitch by an octave ("transforming"), but definitely not both in the same direction.

mdgood:
@mscuthbert We generally have been holding off on graphical issues like alternate representations of octave symbols for later work on specifying styles. These issues are focusing on semantics.

However, it appears that MusicXML 3.1 does not let you specify the symbol used in octave-shift elements either. That might be a good issue to add for MusicXML 3.2.

notator:
@mscuthbert  said:

> I think this is incorrect: "8va: raises both written pitch and sounding pitch by one octave." 

I was trying to get back to the simple, intuitive meaning of an 8va, regardless of how it might be defined in any current standards.
Consider the following:
Instrumentalists memorise _fingerings_ in relation to particular _graphics_. For example, a clarinettist will use the same fingering for a _written C4_, regardless of the instrument in use. If its a Bb-Clarinet, the sounding pitch will be Bb3. If its an A-Clarinet, the sounding pitch will be A3. Transposed parts exist so that instrumentalists can use the same _fingering per graphic_, regardless of the physical size of the instrument being used. That means that players neither have to re-learn the relation between the written pitch (graphic) and the fingering, nor transpose in their heads. (I used to be a horn player...)
Now consider what happens if an 8va sign is added above the written C4. The player now has to use the fingering for C5. In other words, the player regards the graphic as being a _written C5_ (sounding Bb4 or A4, depending on the instrument in use). So adding an 8va above an existing written pitch "raises both written pitch and sounding pitch by one octave."
Maybe we are talking at cross purposes... Hope the above helps clear things up.

adrianholovaty:
@notator: Alas, this pull request is not the place to relitigate the definition of pitch. I've attempted to work around the wording issue in [this followup commit](https://github.com/w3c/mnx/pull/158/commits/df00339e3ad68fe43369c515a138708a109353c1).

notator:
@adrianholovaty Thanks. Okay. The new paragraph above the individual 8va explanations solves the problem by making it clear that its the _score_ rendering that is meant, not the _sound_ rendering or even the rendering in the XML. (You've simply avoided saying anything about the XML or _sound_ rendering.)

As I said above, I still think [§4.12](https://w3c.github.io/mnx/specification/common/#written-pitches) is very confused. Briefly:
Paragraph 1 is a circular definition of "written pitch" as a sound. Definitions need to start from simpler definitions. How do you define "concert pitch instrument" without mentioning "written pitch"?
Paragraph 2 seems to be saying that "written pitch" is, after all, a graphic. How else can it be different from the sound an instrument makes? This paragraph is also circular. It relies on a definition of "transposing instrument".
Paragraph 3: _Obviously_ a graphical marking can have no effect on a sound (that would be a type mismatch). This paragraph is confusing, in that it first says that ottava markings have no effect on a note's written pitch, and then gives an example of them doing just that!

Apropos: Note that [the above thoughts](https://github.com/w3c/mnx/pull/158#issuecomment-526605772) mean that I now agree with the co-chair (and MusicXML) that ottava markings should not be `directions`. They are simply part of the graphic representation of pitches. This is a change from my position in #152.

Shall I open a new issue about defining "written pitch", or would you like to do that? I'm not sure whether you would prefer simply to revise §4.12, or to create a new "basic definitions" section in the spec.

mdgood:
This looks good to me.

PR #159, opened 2019-09-24
Add 'Slurs' section to "By Example" and fill in spec

adrianholovaty:
In this pull request, I've added a "Slurs" section to MNX-Common By Example page, and I've filled in the "Slurs" section of the spec.

To view the By Example page as it stands on this branch:
https://cdn.githubraw.com/w3c/mnx/slurs/by-example/index.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/159.html" title="Last updated on Oct 8, 2019, 2:10 PM UTC (9fb2166)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/159/71ac02d...9fb2166.html" title="Last updated on Oct 8, 2019, 2:10 PM UTC (9fb2166)">Diff</a>


specification/common/index.bs
```diff
@@ -2693,6 +2693,18 @@ location given for the start of the spanning direction.
 
 <h4 id="the-slur-element">The <dfn element><code>slur</code></dfn> element</h4>
 <section dfn-for="slur">
+
+The <code>end</code> attribute specifies the <a>measure location</a> of the
```
mdgood: I have two suggestions here:

1. I think we need to add an equivalent to MusicXML's line-type attribute here to specify solid, dashed, or dotted lines. Unlike most of MusicXML's attributes for the slur element, this one feels semantic to me. Might we want to encode the semantics though instead of the end results? For instance, a dashed slur may be dashed because it happens one time only through a repeated section, or because it is an editorial addition.
2. I think we also need a way to specify the side for the endpoint of the slur, perhaps with an end-side attribute. By default it would be the same value as the side attribute. However in some cases the sides differ, such as slurs that cross staves in a keyboard part.

At some point we want to handle slurs with multiple inflection points. I think that is more of a presentation issue than a semantic issue so we would not need to add this now.

mdgood: More fundamentally, I do not think that a measure location is sufficient for indicating either the start or end point of a slur. This is not sufficiently precise for contrapuntal music with multiple voices in one part or staff. 

MusicXML has slurs specified at the note level. In MNX-Common I think the starting and end points need to be specified at either the event or note level.

This leads me to think that the current MNX-Common spec has the slur element in the wrong place. Instead of being a spanning direction, it seems more appropriate as liaison content, which use notes for both the starting and end points.

Perhaps we may want to close this pull request and add a new issue to discuss the more fundamental move of slurs from spanning direction to liaison content.

cecilios:
The proposal for slurs (#159) and octave-shifts (#158) instead of specifying spanners as a start/end pair of direction elements, only uses a start direction and the end point is specified as a measure location.

To my knowledge, there was no problem with MusicXML approach and, specifying the end point as an attribute in the spanner start element has, IMO, some drawbacks and no gains.

Cons:
- It is not posible to reuse parsing code from MusicXML as well as other code to deal with spanner directions.
- In score edition, inserting or removing notes between the start and the end points of a spanner was very easy in MusicXML model as it was not necessary to deal with spanner directions that could apply to those notes. But in the proposed MNX model, when inserting or deleting a note, it is now necessary to find all applicable spanner directions and to modify its end attribute for taking into account the added/removed time shift.

Pros:
- any?

I would appreciate an explanation of the gains of the proposed approach or, at least, the reasons to move away from the MusicXML way of doing it. Thank you.

cecilios:
After re-thinking several times on the reasons to move away from the MusicXML model I think that now I understand the benefits. I didn't realize that in MNX is only one direction per measure, and each contained direction-type has a start and end location. But in MusicXML model there can be many directions per measure and nothing prevents many consecutive directions.

MNX proposal removes the need to insert a direction to specify the end of a spanner and enforces the restriction of not alowing two consecutive direction elements.

Therefore, nothing to object. Please forget my previous comment. Sorry for the noise.

adrianholovaty:
I'm closing this because we've decided slurs are likely better represented as "liaison content." See the new issue #167 — please chime in with thoughts there. (Note we'll likely still be able to use much of the work from this particular pull request.)

adrianholovaty:
(Note this has now been implemented, in a different way, in pull request #169.)

Issue #167, opened 2019-10-28
Change slurs to be liaison content


adrianholovaty:
At the moment, the MNX-Common spec treats slurs as [spanning directions](https://w3c.github.io/mnx/specification/common/#common-spanning-directions), which led to the pull request #159 implementation. But we've come to realize slurs would be treated better as [liaison content](https://w3c.github.io/mnx/specification/common/#common-liaison-content).

See [this comment](https://github.com/w3c/mnx/pull/159#discussion_r328227208) from @mdgood for more context.

The difference is:

* Spanning directions are defined from one metrical position to another. For example: "This slur spans from a quarter note into bar 2 until three quarter notes into bar 2."
* Liaison content is defined from one _specific_ note/event to another.

Any objections or further thoughts regarding making slurs be liaison content?

clnoel:
The thing about slurs is they are not only on specified start and end notes, they also affect only particular notes in between. For example if you have the slur on a set of down-stem notes, the up stem notes (or notes in the other staff) are not necessarily affected. Does liaison content cover that?

adrianholovaty:
@clnoel I believe that wouldn't be a problem. In your example, the up-stem notes would live in a different voice (a different `<sequence>` in MNX-Common terminology). Liaison content goes from one specific note (which lives in a specific `<sequence>`) to another specific note (which lives in a specific `<sequence>`). Same goes for notes in the other staff. Counter-examples are welcome, of course!

clnoel:
The only counter I can come up with is a slur that spans several measures. The notes in the first and last measures that belong in the slur would be well defined by sequence and liaison specification, but not the notes in the middle measures. If we do this, we need to make a way to optionally specify one or more "midpoints" not just an end point. 

That might be a general requirement for liaison content. Arpeggios could be very well defined this way, and so could slides and bends. Multi-note ties could also benefit from this, so that it is considered a single element instead of a tie that ends on one of the middle notes and then restarts to go on to the next one. (That reflects how its played, and is good for figuring actual note duration fro playback, and display can still stop at each midpoint note.)

adrianholovaty:
Pull request #169 addresses this issue.

clnoel:
While I have no objection to improving the slur specification as written here, I would like to call out that my issue is not addressed by this. I've pulled the following example from the dorico website http://www.rpmseattle.com/of_note/how-to-work-with-slurs-in-dorico-write-mode/

[...]

PR #169, opened 2019-11-12
Implement <slur> as liaison content

adrianholovaty:
In this pull request, I've added a "Slurs" section to MNX-Common By Example page, and I've filled in the "Slurs" section of the spec.

As opposed to the [previous pull request](https://github.com/w3c/mnx/pull/159), this one defines a slur as "liaison content," per various discussion.

To view the By Example page as it stands on this branch:

https://cdn.githubraw.com/w3c/mnx/slurs-liaison/by-example/index.html

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/169.html" title="Last updated on Jan 21, 2020, 4:50 PM UTC (52e2bac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/169/4931ab4...52e2bac.html" title="Last updated on Jan 21, 2020, 4:50 PM UTC (52e2bac)">Diff</a>


adrianholovaty:
This addresses issue #167.

mscuthbert:
For a slur starting or ending on a chord -- must the slur go on the first note element (and the last note element)? What about multiple voices that begin together but then diverge -- are the notes in the middle covered by the slur or not?  (and if the default is the one, how will the other be specified).  I think that it's important that the specification not just give where a slur begins and ends but make it possible for playback and analysis software to easily discern if a note is covered by a slur or not.  (something that is often ambiguous even in standard notation)

I'm not 100% sure whether slurs are spanners or liasions -- I can really see arguments either way!

Great work so far!

mscuthbert:
Must the end element be a note?  I have seen slurs that end on rests as well, and in contemporary music, both slurs and ties that that connect to nothing as a sort of l.v. or with an ornament

adrianholovaty:
@mscuthbert wrote:
> For a slur starting or ending on a chord -- must the slur go on the first note element (and the last note element)?

Thanks for bringing this up. I've just added another example — “Slurs (for chords)” — in this branch. You can [view it here](https://cdn.githubraw.com/w3c/mnx/slurs-liaison/by-example/index.html). The upshot is that the `<slur>` gets added to the first `<note>` in the `<event>`. However, something about that smells weird to me, to be honest, and it makes me think a slur should link to an `<event>` instead of a `<note>` (more on this below)...

> What about multiple voices that begin together but then diverge -- are the notes in the middle covered by the slur or not? (and if the default is the one, how will the other be specified).

Mind posting an image that shows exactly what you mean? It doesn't need to be from published music; something contrived is totally fine. :)

> I think that it's important that the specification not just give where a slur begins and ends but make it possible for playback and analysis software to easily discern if a note is covered by a slur or not. (something that is often ambiguous even in standard notation)

Agreed. The solution here might be to provide some suggested algorithms for crawling through an MNX-Common file and determining which notes are covered by a given slur.

> Must the end element be a note? I have seen slurs that end on rests as well, and in contemporary music, both slurs and ties that that connect to nothing as a sort of l.v. or with an ornament

This is a great point. The spec says "liaison content" links to a `<note>` — so the direct answer to your question is "yes, it must be a note." Your counterexample of a slur ending on a rest makes me think that the slur's `target` should point at an `<event>` rather than a `<note>`. This would cover the case of a slur ending on a rest. I'll bring this up in our biweekly co-chairs meeting later today.

bhamblok:
@adrianholovaty I would like to provide you an alternative technique to write down the examples source-code in the examples-pages.
When you encapsulate the sample-xml within a script-tag as follows, you don't have to transcribe the "<" and ">"-symbols into html-entities like ```&lt;``` ```&gt;``` which improves the readability of the source-code ;-)
```
<script type="application/xml" style="display: block; white-space: pre;">
  <mnx>
     <...>
  </mnx>
</script>
```

adrianholovaty:
@bhamblok: Ooooh, I didn't know about that — thanks! What's the browser support like?

bhamblok:
@adrianholovaty not sure... but, this is really basic usage of a script-tag and 2 css-props which are supported up to IE6 :-D

adrianholovaty:
@bhamblok: Now that I think about it, I don't think this is feasible for us to use — because we embed some markup in the examples that highlights specific sections (`<div class="diff">` and `<div class="nodiff">`). If there's an elegant way to do this with `<script type="application/xml">`, let me know.

bhamblok:
@adrianholovaty OK ... right ... I see.
But I think it might be worth experimenting to simply exchange
```<div class="diff> or <div class="nodiff">```
with
```<script type="application/xml" class="diff"> or <script type="application/xml" class="nodiff">```
And where you add the css-property "white-space: pre;" to the classes ".diff" and ".nodiff"...
I hope this works.

adrianholovaty:
@bhamblok Ah, yes, I think that will work! I'm working on the change now — expect a commit soon. :)

adrianholovaty:
@bhamblok That's now committed in 8dd67a711. Thanks again for the suggestion!

cecilios:
Before closing this issue I think that it is important to clarify if it is mandatory that start and end points of an slur be on the same sequence.

This was suggested in issue 167 (https://github.com/w3c/mnx/issues/167#issuecomment-549413786) but open questions remained ( https://github.com/w3c/mnx/issues/167#issuecomment-549425188)

Later, in the comments to this PR it was also raised the question of how to determine the notes covered by an slur, and it was pointed that (see https://github.com/w3c/mnx/pull/169#issuecomment-558638486):
> Agreed. The solution here might be to provide some suggested algorithms for crawling through an MNX-Common file and determining which notes are covered by a given slur.

But I cannot find a reference to the constrain of start and end points of an slur be on the same sequence, or about any suggested algorithms for determining the notes affected by the slur. My apologies if a missed something.

clnoel:
@cecilios I don't think we can do that. To my knowledge, sequences are within-measure only. Slurs can go across measures. If we were to identify a way to make sequences from different measures belong in the same "meta-sequence", we could make that restriction, and it would also solve the problem of identifying which notes in intervening measures belong to the slur.

--clnoel

cecilios:
@clnoel Yes, you are right, sequences are within-measure only. I did not explain myself right , I tried to refer to sequences with the same voice attribute. But I'm not sure if forcing start and end notes of a slur having to be on the same voice is an acceptable constrain.

cecilios:
Another issue in relation to this PR is the ordering of `<note>` and `<slur>` elements in the "Slurs (for chords)" example. According to current content model for the `<event>` element, slurs must go always after the notes.

The example for chords seems to suggest that for chords the slur must be defined before the notes, but I do not see the need for this, as according to current definition they always go attached to the event, and start/end notes, if required, are  can be specified using the start-note and end-note attributes.

Issue #165, opened 2019-10-28
Specify how/whether barlines are encoded


adrianholovaty:
The MNX-Common spec should explicitly state whether barlines must be encoded. A couple of options:

* Treat them as opt-out: i.e., the document assumes barlines exist and are rendered, without needing to explicitly specify them, unless you turn them off.
* Treat them as opt-in: i.e., the document assumes barlines are _not_ rendered unless you explicitly specify them.

My inclination is to treat them as opt-out, given that the vast majority of CWMN will use barlines in the common way. This will make MNX-Common encodings simpler and will cause fewer mistakes (e.g., people forgetting to encode the barlines).

Note there are a few types of barlines I think this "opt-out" should apply to:

* Regular barlines (e.g., in the middle of a system)
* Systemic barlines (e.g., at the start of a system, connecting staves)
* Final barline (e.g., at the end of the piece)

For out-of-the-ordinary barlines, such as double barlines, it seems clear that those should be opt-in.

cecilios:
Perhaps this issue is related to #61. And also to the very old issue of the role of the `<measure>` container: whether it represents a measure or it is just a helper container (equivalent to a html div).

It is not my intention to re-open old issues, but just to recall that it was decided that `<measure>` will be a _rich_ container representing a measure, as in MusicXML, and not just an auxiliary helper container. Therefore, if `<measure>` represents a measure, the most logical conclusion is to accept that a barline is by default implied at the end of the measure. Thus, the opt-out seems to be the only coherent option for regular barlines.

As to final barlines, probably it is better to use a fixed default (regular barline) instead of a default that changes depending on the position of the measure element (regular or final barline). Both options are acceptable to me. But if final barline is implied by last measure container, then for use cases such as snippet scores there should be a way of overriding this and placing, instead, a regular barline.

As to the systemic barline, I'm not sure if it is implied by the measure container or by the fact of laying out a score with metric. I think of it more as a global option for score layout or a local direction, that something implied by a measure container, but that's just my feeling and I can not give supporting arguments.

clnoel:
System barlines should be part of the layout style, however we end up specifying that, not part of the measure container. Different layouts will have the system barlines in between different measures, or might only be extracting one part, or might be a snippet that doesn't need a system barline at all.

adrianholovaty:
Pull request #171 addresses this.

adrianholovaty:
This was addressed by commit 1f6b11c.

PR #171, opened 2019-11-12
Specify barline defaults

adrianholovaty:
This updates the MNX-Common spec to give the defaults for barlines.

It also updates the MNX-Common By Example page to make the markup examples match the screenshots, with respect to the final barline. (Some examples use a proper final barline, while others use a "normal" barline.)

This addresses issue #165.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/171.html" title="Last updated on Nov 14, 2019, 2:18 PM UTC (03866ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/171/4b6701a...03866ae.html" title="Last updated on Nov 14, 2019, 2:18 PM UTC (03866ae)">Diff</a>

### DIFF
--- a/by-example/index.html
+++ b/by-example/index.html
@@ -123,10 +123,10 @@ a {
 
     <p><strong>IMPORTANT NOTE: The MNX-Common spec is not yet ready for implementation.</strong> MNX-Common is still a work in progress, and we hope you’ll get involved with its development. Does anything look weird in these examples? Hard to parse? Confusing? Potentially ambiguous? We want to hear about it. Make sure you’ve joined the <a href="https://www.w3.org/community/music-notation/">W3C Music Notation Community Group</a>, and post your ideas to <a href="https://github.com/w3c/mnx/issues">our GitHub issue tracker</a>.</p>
 
-    <p>Some notes about these examples:</p>
+    <p>Some notes about these MNX-Common examples:</p>
     <ul>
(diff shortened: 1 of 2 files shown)


mscuthbert:
Is the end barline to be called `light-heavy` or `final`?  It looks like a conflict in the text.  The former keeps musicxml compatibility -- the latter is what I would've named it originally if I were emperor (along with `double` instead of `light-light`)

adrianholovaty:
@mscuthbert That's a great suggestion. I've just opened #172 to track that decision, as it's slightly different than what this specific pull request deals with.

For the record, the barline types _are_ indeed consistent in this pull request. You may be seeing the `barline="final"` in the diff, but that's a part of the file that's been removed.

mscuthbert:
> For the record, the barline types _are_ indeed consistent in this pull request. You may be seeing the `barline="final"` in the diff, but that's a part of the file that's been removed.

I saw that, but I was also talking about the prose part that said the default is all regular barlines except the last which is `final`.  But if #172 is passed then no need.

adrianholovaty:
> I saw that, but I was also talking about the prose part that said the default is all regular barlines except the last which is `final`.

Ah, I see what you mean! That prose part is referring to the generic concept of final barlines, but the double-quote around it makes it seem More Important than it actually is. :) Will tweak this to make it less misleading.

PR #186, opened 2020-06-09
Add repeat and jump examples to /by-example/

adrianholovaty:
In this branch, I've added the following examples to "MNX-Common By Example":

* Repeats
* Repeats (with implied start repeat)
* Repeats (with alternate endings)
* Jumps (Dal Segno)
* Jumps (D.S. al Fine)

This pull request does not include any changes to the spec. The intent here is to demonstrate a proposed syntax for repeats and jumps, get feedback on it, iterate, etc., before spending time on updating the spec. I will update the spec when we've agreed on the approach in these examples.

You can view the new examples [via this link](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html) (see the last five examples in the list).

Some notes:

* I haven't implemented every single jump. E.g., this is missing Da Coda. I deliberately chose a simple jump (Dal Segno) and a complex jump (D.S. al Fine) to demonstrate the general approach. Other jumps would work in the same way.
* The approach to jumps deliberately doesn't specify a "bar number to jump to." It's the responsibility of the MNX-Common consumer application to interpret the jumps. We can eventually add a section to the spec that provides a suggested algorithm for this. For example: "If you encounter `<jump type="dsalfine">`, traverse the measures backwards until you find `<segno>` and play from the measure containing `<segno>` until the measure containing `<fine>`, inclusive.
* There also needs to be a way to specify the interaction between jumps and repeats, to be able to say "After the Dal Segno, don't repeat when you reach the repeat bars." MusicXML does this with the [time-only element](http://usermanuals.musicxml.com/MusicXML/MusicXML.htm#EL-MusicXML-sound_1.htm). In MNX-Common, I expect this would be an attribute of the `<jump>` element.

Feedback welcome!


notator:
Great to be talking code! :-)

[Repeats](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats1) and [Repeats (with implied start repeat)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats2):
I think you could simply add a `times` attribute to the `<repeat type="end" ` barline, to advise on the number of times to hit the barline during a performance. The default would be `times="2"`.
It would be up to the consuming application whether/how to display that info in the score (just as its up to the app how to display the barline itself).
In other words, you could have written the following in both examples:
```
<repeat type="end" times="3"/>
```
##### Note that its very easy for a performing application to keep track of the number of times a barline has been hit.
---
[Repeats (with alternate endings)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats3):
I think the example is a bit too simple. Endings are not always just one measure long, and you need to show how multi-measure endings are going to work.
I think an `<ending>` is actually a kind of jump back to the most recently hit repeat barline. So it should be possible to put `<ending>` elements in any `<measure>` preceding  a `<repeat type="end">` barline.
I think there may be a general principle here, that would also be useful for segnos: `<ending>` and `<fine>` elements are conditional jumps. Maybe the `<ending>` element could be replaced by a `<jump type="ending">1.</jump>`. (see below)

---
[Jumps (D.S. al fine)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#jumps2)
If all `<fine>` elements are conditional on a preceding `<segno>` having been activated (by a `<jump type="dalsegno">D.S. al Fine</jump>`), then you don't need the `"dsalfine"` jump type.

adrianholovaty:
@notator @clnoel Thanks very much for the comments! Based on your feedback, I've changed/expanded the examples in this pull request. The changes are:

* `<ending>` now uses separate `start` and `end` elements, and I've updated the [first alternate endings](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats3) example.
* I've added [an example](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats4) with more complex alternate endings (across multiple bars and with multiple ending numbers).
* I've added [an example](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats5) of a "4x" ending.
* I've removed the text within the `<jump>` elements. Previously this was intended to be a place to write the displayed/graphical text, but we [decided in our last co-chairs meeting](https://www.w3.org/community/music-notation/2020/06/09/co-chair-meeting-minutes-june-8-2020/) that this wouldn't be expressive enough for the case of "D.S. al [CODA SIGN GRAPHIC]". The goal is to come up with something more holistic across all of MNX, for the general pattern of "include some text which might in turn include some graphical elements."

You can see the latest and greatest version of the "by example" page in this pull request [here](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html). More comments welcome!

notator:
@adrianholovaty Thanks for the update and new examples.
The first three repeat examples and the dalsegno examples look fine. Here are some thoughts about [Repeats (with alternate endings, simple)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats3) and [Repeats (with alternate endings, advanced)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats4):

Using multiple ending numbers (in the `<ending> number ` attribute) looks okay, but I think the `<ending type="stop"/>` is redundant.
At least in these examples, endings _always_ either stop at a `<repeat type="end"/>` or never stop at all.
That leaves only one type of ending, so `<ending>` doesn't need a type attribute.
In other words, I think [Repeats (with alternate endings, simple)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats3) could be rewritten using
```
<global>
    <measure>
        <directions>
            <time signature="4/4"/>
            <repeat type="start"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <ending number="1"/>
            <repeat type="end"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <ending number="2"/>
            <repeat type="end"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <ending number="3"/>
        </directions>
    </measure>
</global>
```
(To make it clearer what's going on, I've put the `<ending>`s before the `<repeat>`s in the `<directions>` elements above. The order doesn't actually make any functional difference.)

[Repeats (with alternate endings, advanced)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats4) could be rewritten using
```
<global>
    <measure>
        <directions>
            <time signature="3/4"/>
            <repeat type="start"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <ending number="1,3"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <repeat type="end"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <ending number="2"/>
        </directions>
    </measure>
    <measure>
        <directions>
            <repeat type="end"/>
        </directions>
    </measure>
    <measure/>
</global>
```
Not having to keep track of two types of `<ending>` simplifies the performing algorithm. In fact, it then looks very similar to the algorithm for performing the dalsegnos, and just means managing the state of the endings according to the number of times the repeat-end barlines have been hit.
(This also the way human performers interpret the score.)

---

I think ending-brackets should be open and one bar long by default (when they don't end), and _always_ be closed where they do end. So:
- The 3. bracket in the graphic for [Repeats (with alternate endings, simple)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats3), would be open and one bar long even if there were more bars added to the end of the example.
- The 2. bracket in the graphic for [Repeats (with alternate endings, advanced)](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats4) should have a closing vertical at the end of measure 5.



Are there any more complex examples that haven't been covered?

mdgood:
I think the advanced alternate endings example could be more helpful if it reflected a more common situation of the earlier ending being 1,2 and the latter ending being 3. This would eliminate the issue of the repeat ending always coinciding with a backward repeat in these examples. 

It would also allow us to show how discontinued endings work, as the 3rd ending would not have a downstroke at the end, nor would its endpoint be something you could always determine automatically. In MusicXML this is represented by a "discontinue" type value for the ending element. This is something that could also be addressed in the simple example, which uses a "stop" type where MusicXML uses "discontinue" and thus is losing information.

adrianholovaty:
@mdgood Thanks! All makes sense. I've just updated the advanced altendings example:

* Bars 2-3 now use endings "1,2" instead of "1,3", and bars 4-5 now use ending "3" instead of "2"
* The MusicXML now uses `type="discontinue"` for the final ending.
* Removed the now- unnecessary end-repeat bar at the end of bar 5.

[Here's the link where you can see it in action](https://cdn.githubraw.com/w3c/mnx/repeats-jumps/by-example/index.html#repeats4).

@notator That's a tempting idea to allow people to leave out the `<ending type="stop">`, but it is indeed necessary for final endings (like ending 3 in the newest version of the advanced altendings example), so that we know how long that final ending lasts. I'm inclined to favor consistency — always require a start/stop — as opposed to supporting an implicit `<ending type="stop">`.

notator:
@mdgood, @adrianholovaty  Thanks for the revision. Understood. okay!

Issue #174, opened 2020-03-31
Repeats and alternate endings


adrianholovaty:
The MNX-Common spec [hasn't yet implemented repeats](https://w3c.github.io/mnx/specification/common/#the-repeat-element) or [alternate endings](https://w3c.github.io/mnx/specification/common/#the-ending-element). Here's a dump of examples and questions to ponder.

First, some notes/observations:

* We need to encode the "start repeat" and "end repeat" locations, which might be in separate bars.
* Semantically, "start repeat" means "the bar that's returned-to after the end-repeat bar is performed." "End repeat" means "the last bar that's performed before jumping to the start-repeat bar."
* "Start repeat" might be missing/implied (see Example 2 below).
* We need to encode the number of times to repeat (see Example 3 below).
* In cases with more than two alternate endings, engraving practices differ (see Examples 5 and 6 below).
* A single bar might contain more than one ending number (see Example 7 below).

Some examples, so we have concrete things to consider:

**Example 1: Simple repeat**

<img width="496" alt="Screen Shot 2020-03-31 at 11 43 43 AM" src="https://user-images.githubusercontent.com/180401/78012174-e4b10c00-7344-11ea-8eaf-86b238d7eb2f.png">

**Example 2: Simple repeat with implied start-repeat**

<img width="497" alt="Screen Shot 2020-03-31 at 11 49 46 AM" src="https://user-images.githubusercontent.com/180401/78013226-7a00d000-7346-11ea-9c80-88008c54dad1.png">

**Example 3: Simple repeat with 4x**

<img width="499" alt="Screen Shot 2020-03-31 at 11 50 02 AM" src="https://user-images.githubusercontent.com/180401/78013245-7f5e1a80-7346-11ea-8cc1-14cdab0e5886.png">

**Example 4: Simple alternate endings**

<img width="588" alt="Screen Shot 2020-03-31 at 11 51 30 AM" src="https://user-images.githubusercontent.com/180401/78013295-8f75fa00-7346-11ea-9b29-1097d3f23bd0.png">

**Example 5: Three alternate endings (two end-repeat bars)**

<img width="722" alt="Screen Shot 2020-03-31 at 11 52 09 AM" src="https://user-images.githubusercontent.com/180401/78013321-98ff6200-7346-11ea-8516-91c021bae420.png">

**Example 6: Three alternate endings (one end-repeat bar)**

This is an alternate approach to Example 5. Only one end-repeat bar is visible.

[...]

PR #190, opened 2020-07-21
Add repeats and jumps to spec

adrianholovaty:
This pull request adds the following definitions to the MNX-Common spec:

* `<ending>`
* `<repeat>`
* `<coda>`
* `<segno>`
* `<jump>`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/190.html" title="Last updated on Jul 21, 2020, 9:59 AM UTC (7f1ef51)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/190/25909b9...7f1ef51.html" title="Last updated on Jul 21, 2020, 9:59 AM UTC (7f1ef51)">Diff</a>

### DIFF
--- a/specification/common/index.bs
+++ b/specification/common/index.bs
@@ -2770,12 +2770,91 @@ means a repeat ending at the end of this measure.
 
 <h4 id="the-coda-element">The <dfn element><code>coda</code></dfn> element</h4>
 <section dfn-for="coda">
-  <em>Needs migration from MusicXML. May need to distinguish musical form use from textual reference to the symbol.</em>
+  <dl class="def">
+    <dt><a>Contexts</a>:</dt>
+    <dd>Direction content</dd>
+    <dt><a>Content Model</a>:</dt>
+    <dd>None</dd>
+    <dt><a>Attributes</a>:</dt>
+    <dd>None</dd>
+  </dl>
+
+The <{coda}> element describes a coda symbol.
+
+This element applies to the containing measure. The coda symbol is traditionally
+rendered at the start of its measure.
 </section>
 
 <h4 id="the-segno-element">The <dfn element><code>segno</code></dfn> element</h4>
 <section dfn-for="segno">
-  <em>Needs migration from MusicXML. May need to distinguish musical form use from textual reference to the symbol.</em>
+  <dl class="def">
+    <dt><a>Contexts</a>:</dt>
+    <dd>Direction content</dd>
+    <dt><a>Content Model</a>:</dt>
+    <dd>None</dd>
+    <dt><a>Attributes</a>:</dt>
+    <dd>None</dd>
+  </dl>
+
+The <{segno}> element describes a segno symbol.
+
+This element applies to the containing measure. The segno symbol is traditionally
+rendered at the start of its measure.
+</section>
+
+<h4 id="the-fine-element">The <dfn element><code>fine</code></dfn> element</h4>
+<section dfn-for="fine">
+  <dl class="def">
+    <dt><a>Contexts</a>:</dt>
+    <dd>Direction content</dd>
+    <dt><a>Content Model</a>:</dt>
+    <dd>None</dd>
+    <dt><a>Attributes</a>:</dt>
+    <dd>None</dd>
+  </dl>
+
+The <{fine}> element describes a fine symbol.
+
+This element applies to the containing measure. The fine symbol is traditionally
+rendered at the end of its measure.
+</section>
+
+<h4 id="the-jump-element">The <dfn element><code>jump</code></dfn> element</h4>
+<section dfn-for="jump">
+  <dl class="def">
+    <dt><a>Contexts</a>:</dt>
+    <dd>Direction content</dd>
+    <dt><a>Content Model</a>:</dt>
+    <dd>None</dd>
+    <dt><a>Attributes</a>:</dt>
+    <dd><{jump/type}> - (required) see choices below</dd>
+  </dl>
+
+The <{jump}> element describes a musical jump: a direction to continue
+performance at some other measure.
+
+The <dfn element-attr dfn-for="jump">type</dfn> attribute specifies the type of jump:
+
+<dl dfn-for="jump/type">
+  <dt><dfn attr-value><code>segno</code></dfn></dt>
+  <dd>"Dal segno": Jump to the <{segno}></dd>
+  <dt><dfn attr-value><code>coda</code></dfn></dt>
+  <dd>"To coda": Jump to the <{coda}></dd>
+  <dt><dfn attr-value><code>capo</code></dfn></dt>
+  <dd>"Da capo": Jump to the first measure</dd>
+  <dt><dfn attr-value><code>dcalfine</code></dfn></dt>
+  <dd>"D.C. al fine": Jump to the first measure, then perform until the measure that contains <{fine}> (inclusive)</dd>
+  <dt><dfn attr-value><code>dcalcoda</code></dfn></dt>
+  <dd>"D.C. al coda": Jump to the first measure, then perform until the measure that contains <{jump} type="coda"> (inclusive), then jump to the coda</dd>
+  <dt><dfn attr-value><code>dsalfine</code></dfn></dt>
+  <dd>"D.S. al fine": Jump to the <{segno}>, then perform until the measure that contains <{fine}> (inclusive)</dd>
+  <dt><dfn attr-value><code>dsalcoda</code></dfn></dt>
+  <dd>"D.S. al coda": Jump to the <{segno}>, then perform until the measure that contains <{jump} type="coda"> (inclusive), then jump to the coda</dd>
+</dl>
+
+This element applies to the containing measure and is intended to be
+followed after the measure's music. The direction is traditionally rendered
+at the end of its measure.
 </section>
 
 <h4 id="the-harmony-element">The <dfn element><code>harmony</code></dfn> element</h4>
--- a/specification/common/index.bs
+++ b/specification/common/index.bs
@@ -2720,14 +2720,52 @@ consistent with describing musical expression.
 
 <h4 id="the-ending-element">The <dfn element><code>ending</code></dfn> element</h4>
 <section dfn-for="ending">
-  <em>Needs migration from MusicXML. Applies from the start of the containing measure (i.e. does not belong to a &lt;barline&gt;)</em>
+  <dl class="def">
+    <dt><a>Contexts</a>:</dt>
+    <dd>Direction content</dd>
+    <dt><a>Content Model</a>:</dt>
+    <dd>None</dd>
+    <dt><a>Attributes</a>:</dt>
+    <dd><{ending/type}> - (required) <code>start</code>, <code>stop</code> or <code>discontinue</code></dd>
+    <dd><{ending/number}> - (required for <code>type="start"</code>) a comma-separated list of integers representing the ending numbers</dd>
+  </dl>
+
+The <{ending}> element describes an alternate ending, also known as a "volta."
+
+Each ending in a document must include two <{ending}> elements: one with
+<code>type="start"</code> and another with either <code>type="stop"</code>
+or <code>type="discontinue"</code>.
+
+Use <code>type="stop"</code> when the ending concludes with a downward jog,
+as is typical for first endings. Use <code>type="discontinue"</code> when
+there is no downward jog, as is typical for second endings that do not
+conclude a piece.
+
+The <code>number</code> attribute uses one-indexed numbers, not
+zero-indexed numbers. For example, the first ending has <code>number="1"</code>.
+
+This element applies to the containing measure. Thus, <code>type="start"</code>
+means an ending starting from the beginning of this measure, and <code>type="stop"</code>
+means an ending that stops at the end of this measure.
 </section>
 
 <h4 id="the-repeat-element">The <dfn element><code>repeat</code></dfn> element</h4>
 <section dfn-for="repeat">
-  <em>Needs migration from MusicXML. Applies to the containing measure (i.e. does not belong to a &lt;barline&gt;).
-      Thus, "start" means a repeat starting from the beginning of this measure; "end" means a repeat ending at the end
-      of this measure.</em>
+  <dl class="def">
+    <dt><a>Contexts</a>:</dt>
+    <dd>Direction content</dd>
+    <dt><a>Content Model</a>:</dt>
+    <dd>None</dd>
+    <dt><a>Attributes</a>:</dt>
+    <dd><{repeat/type}> - (required) either <code>start</code> or <code>end</code></dd>
+    <dd><{repeat/times}> - (optional) the number of times the music within the repeat should be performed. Default is 2.</dd>
+  </dl>
+
+The <{repeat}> element describes a repeat, traditionally rendered as a repeat barline.
+
+This element applies to the containing measure. Thus, <code>type="start"</code>
+means a repeat starting from the beginning of this measure, and <code>type="end"</code>
+means a repeat ending at the end of this measure.
 </section>
 
 <h4 id="the-coda-element">The <dfn element><code>coda</code></dfn> element</h4>


clnoel:
Now that I'm looking at this after a break, I see the following:
- Since these are "directions" we probably need to allow "Attribute Groups: Direction attributes"
- - Default "Measure location" for endings "Stop" and "Discontinue" and repeat "end" is actually the end of the measure (not locatoin 0).
- We need to be able to handle more than one `segno` and `coda` in a piece without confounding the `jump` direction. Double-segnos and Double-codas are rare but not unheard of. If that needs to be a new issue/PR, I'm okay with that.
- The `jump` direction needs optional text content (default text depends on `type`)

adrianholovaty:
@clnoel Thanks for giving it a read!

>     * Since these are "directions" we probably need to allow "Attribute Groups: Direction attributes"
> 
>     * * Default "Measure location" for endings "Stop" and "Discontinue" and repeat "end" is actually the end of the measure (not locatoin 0).

I was conflicted on this one. Technically a measure location is meaningless for a repeat, ending or jump; see the sentences beginning with "This element applies to the containing measure" in the diff.

>     * We need to be able to handle more than one `segno` and `coda` in a piece without confounding the `jump` direction. Double-segno, Double-coda, and "Coda II" are rare but not unheard of. If that needs to be a new issue/PR, I'm okay with that.

Yes, great point. I discussed this with Daniel and Michael in our biweekly video chat today, and we came to the conclusion that the best solution would be to allow multiple segnos/codas in the same piece but require unique identifiers to disambiguate. I'll update this pull request with that change...

>     * The `jump` direction needs optional text content (default text depends on `type`)

We're going to punt on this particular issue at the moment, because we need to be able to encode things like "D.S. al [CODA SYMBOL]" (i.e., with embedded graphics). The eventual solution will need to handle that, plus possibly fonts and other stylistic choices. This will likely be addressed by a larger-scale focus on presentation/style.

notator:
Oops, sorry, I originally posted this in  issue #207, but now see that it belongs here.

I'm currently working on converting the [Repeats examples in _MNX by Example_](https://w3c.github.io/mnx/by-example/#repeats1) to SVG using my [MNXtoSVG](https://github.com/notator/MNXtoSVG) application.

According to [Gardner Read's _Music Notation_ (1974), page 227](https://www.amazon.com/Gardner-Read-Notation-Manual-Practice/dp/B00HTJTMGS):

GardnerReadRepeats


[MusicXML repeats](https://www.musicxml.com/tutorial/the-midi-compatible-part/repeats/) always seem to be barline types, which would make the above constructions impossible.
So: a couple of questions for @mdgood:
1. Does MusicXML allow repeat signs to be placed as in the above examples? Maybe I've missed something...
2. If not, is it going to support this in a future version?

Whatever the answers, this is standard practice, so MNX needs to support it, and we need a corresponding example (or two) in _MNX by Example_. The obvious way to do it in MNX would be to provide an optional [`measureLocation` attribute](https://w3c.github.io/mnx/specification/common/#measure-locations) on the `<repeat>` element. This would default to 0 for `<repeat type="start" />` and to the duration of the measure for `<repeat type="end" />`.

See also the comment in [§6.8.6 The repeat element](https://w3c.github.io/mnx/specification/common/#the-repeat-element) in the Draft Spec:
> [The repeat element] _Applies to the containing measure (i.e. does not belong to a `<barline>`). Thus, "start" means a repeat starting from the beginning of this measure; "end" means a repeat ending at the end of this measure_.

mdgood:
@notator The first example is one of the aspects of graphical appearance that is not notated explicitly in MusicXML 3.1, similar to the exact location of cautionary key signatures. Notation applications generally handle this situation automatically without needed guidance at the appearance level from the MusicXML representation. The information is represented semantically, so the beginning of the repeat and the new time and key signatures would all be represented at the start of the second measure. Notation application will then automatically generate  the desired display.

The second example can be represented in MusicXML 3.1 using a `<barline location="middle">` element in the middle of the measure. Alternatively you can have the repeat start a new measure, leaving the measures on either side of the repeat short of the complete duration of the time signature.

notator:
@mdgood Thanks. Understood.
That means we only need an _MNX by Example_ example for mid-measure repeats. The first case can be left for consuming applications to sort out.

For this PR: As explained above, I'd like to propose that the MNX `<repeat>` element be given an optional [`location` attribute](https://w3c.github.io/mnx/specification/common/#measure-locations). This would default to 0 for `<repeat type="start" />` and to the duration of the measure for `<repeat type="end" />`.

Below is how I want to code the following example in MNX. Perhaps you could provide the equivalent MusicXML 3.1 code so that it can be included in _MNX by Example_?
repeat-mid-measure
```
<mnx>
    <global>
        <measure>
            <directions>
                <time signature="3/4"/>
            </directions>
        </measure>
        <measure>
            <directions>
                <repeat type="start" location="1/2"/>
            </directions>
        </measure>
        <measure />
        <measure>
            <directions>
                <repeat type="end" location="1/2"/>
            </directions>
        </measure>
        <measure/>
    </global>
    <part>
        <part-name>Music</part-name>
        <measure>
            <directions>
                <clef sign="G" line="2"/>
            </directions>
            <sequence>
                <event value="/2d">
                    <note pitch="C5"/>
                </event>
            </sequence>
        </measure>
        <measure>
            <sequence>
                <event value="/2">
                    <note pitch="E5"/>
                </event>
                <event value="/4">
                    <note pitch="D5"/>
                </event>
            </sequence>
        </measure>
        <measure>
            <sequence>
                <event value="/2d">
                    <note pitch="G5"/>
                </event>
            </sequence>
        </measure>
        <measure>
            <sequence>
                <event value="/2">
                    <note pitch="G5"/>
                </event>
                <event value="/4">
                    <note pitch="F5"/>
                </event>
            </sequence>
        </measure>
        <measure>
            <sequence>
                <event value="/2d">
                    <note pitch="E5"/>
                </event>
            </sequence>
        </measure>
    </part>
</mnx>
```

adrianholovaty:
@notator I agree it would be good to add an example for that case.

For the actual markup, I'm inclined to prefer the simple approach Michael suggested: "have the repeat start a new measure, leaving the measures on either side of the repeat short of the complete duration of the time signature." So no special markup would be needed for this case.

notator:
Both @clnoel and the co-chair have reservations about the way **Jumps** are currently being coded in _MNX by Example_, and I need some clarification before implementing them in my MNXtoSVG application. 
[The co-chair says](https://github.com/w3c/mnx/pull/190#issuecomment-662046171):
>We're going to punt on this particular issue at the moment, because we need to be able to encode things like "D.S. al [CODA SYMBOL]" (i.e., with embedded graphics). The eventual solution will need to handle that, plus possibly fonts and other stylistic choices. This will likely be addressed by a larger-scale focus on presentation/style.

To answer that, and drawing on @clnoel's comments [above](https://github.com/w3c/mnx/pull/190#issuecomment-661928631) and in https://github.com/w3c/mnx/issues/174#issuecomment-608499861, I'd like to make the following proposal:

There should be two basic elements: `<target>` and `<jump>`, both of which can be defined in `<global><measure><directions>`.
- The `<target>` element is simply a place to jump to (like a "label" in many programming languages).
- The `<jump>` element can perform one of three different actions when encountered in a performance:
   -  continue,
   -  goto somewhere 
   -  stop.
   
The `<jump>` actions will be represented as elements in an ordered list inside the `<jump>` element, each of which is used once (in order) as the `<jump>` is repeatedly executed. The examples below should make this clearer.

The simple _dal Segno_ example from _MNX by Example_:
dalSegno
would be coded like this:
```
<mnx>
    <global>
        <measure>
            <directions>
                <time signature="4/4"/>
            </directions>
        </measure>
        <measure>
            <directions>
                <target id="target1" text="&segno;" />
            </directions>
        </measure>
        <measure/>
        <measure/>
        <measure>
            <directions>
                <jump text="D.S."> 
                    <goto target="#target1" />
                </jump>
            </directions>
        </measure>
    </global>
    ...
</mnx>
```
Remarks:
- I can't find any official HTML token strings for the [segno](https://www.compart.com/en/unicode/U+1D10B) or [coda](https://www.compart.com/en/unicode/U+1D10C) characters apart from the numeric ones, but we could use the HTML pattern anyway in MNX to make legible equivalents. In other words, we could use ("&segno;") for the segno character, and ("&coda;") for the coda character. Possibly, such strings have been defined officially for [all the unicode music symbols](https://www.compart.com/en/unicode/block/U+1D100), but I haven't found them.
- Probably, MNX will eventually be able to contain CSS recommendations for element and attribute classes (also for coloured noteheads etc.). In addition to the simple _segnos_ and _codas_ defined here, it would also be possible for other objects (such as rehearsal numbers) to be targets, and it would be quite possible to define a `<target> class` attribute that would take care of the appropriate styling. However, I think that's a separate issue. Maybe we should start addressing it here. Maybe not. (?)
- `<target>` elements are (like `<repeat type="begin">` placed, by default, at the _beginning_ of the measure in which they are defined.
- `<jump>` elements are (like `<repeat type="end">`) placed, by default, at the _end_ of the measure in which they are defined. Note that there could be both a `<jump>` and a `<repeat type="end">` at the end of a `<measure>`. In that case, the `<repeat>` would always be executed (as many times as it says) _before_ executing the `<jump>`.
- the `target` of a `<goto>` element is a [measure location](https://w3c.github.io/mnx/specification/common/#measure-locations), so could either be an ID (as above) or a measure index (In the above case, the goto would be `<goto target="2:0"` and the target ID would be superfluous.)

The D.S. al Fine example:
dalSegnoAlFine
would be coded like this:
```
<mnx>
    <global>
        <measure>
            <directions>
                <time signature="4/4"/>
            </directions>
        </measure>
        <measure>
            <directions>
                <target id="target1" text="&segno;" />
            </directions>
        </measure>
        <measure>
            <directions>
                <jump text="fine">
                    <continue />
                    <stop /> 
                </jump>
            </directions>
        </measure>
        <measure/>
        <measure>
            <directions>
                <jump text="D.S. al Fine"> 
                    <goto target="#target1" />
                </jump>
            </directions>
        </measure>
    </global>
    ...
</mnx>
```

And a D.S. al Coda (not yet in _MNX by Example_)
dalSegnoAlCoda

would be coded like this:
```
<mnx>
    <global>
        <measure>
            <directions>
                <time signature="4/4"/>
            </directions>
        </measure>
        <measure>
            <directions>
                <target id="target1" text="&segno;" />
            </directions>
        </measure>
        <measure>
            <directions>
                <jump text="al Coda">
                    <continue />
                    <goto target="#target2" /> 
                </jump>
            </directions>
        </measure>
        <measure/>
        <measure>
            <directions>
                <jump text="D.S. al Coda"> 
                    <goto target="#target1" />
                </jump>
            </directions>
        </measure>
        <measure/>
        <measure>
            <directions>
                <target id="target2" text="&coda;" />
            </directions>
        </measure>
    </global>
    ...
</mnx>
```
(This example is like the one in the [Finale documentation](https://usermanuals.finalemusic.com/Finale2012Win/Content/Finale/D_S_D_S_al_Coda.htm). Note that Measure 6 never gets played.)
@clnoel mentions "double-segno", "double-coda", and "Coda II" forms. I think these, and other more complex routes through the score, could also be coded using the above elements. A double-segno would simply be a `<target text="&segno;&segno;">`.

Questions:
- Do we want to allow mid-measure `<target>` and/or `<jump>` elements? I think that would increase their power considerably. It could easily be done (as with `<repeat>`s) by giving them an optional `location` attribute.
- Should target ID reference strings always begin with the '#' character (as in HTML)? That's what the Draft Spec says, but I don't think we've been doing that consistently in _MNX by Example_.
- As it stands, the `text` attributes of both `<target>` and `<jump>` can contain newlines ('\n'). Is that okay?

clnoel:
I approve of the general outline of these jumps and targets. I would prefer to allow mid-measure jumps and targets, even if they are rare.

I think that we don't need the # in the targets, but I'll defer that to someone who is more familiar with general XML/CSS styles.

We should allow newlines. Sometimes space gets crunched and you need to put things on two lines. Although rare, there's no reason to exclude the possibility.

As a side note, I _really like_ the idea of using "fake" HTML special-symbol notation for common musical symbols that end up in text directions, like segno and coda. It solves so many issues that we have to scramble around. And lets different reading applications apply whatever algorithm they need, like changing to their special music font, in order to properly display it.

--Christina

ahankinson:
The hash fragment format (identifiers starting with a `#`) ties an ID in with the URI specifications for identifying sub-parts of a document. Seeing an ID reference start with a hash fragment is effectively shorthand for saying "this ID in the current document" (as opposed to "this ID in some other document on the web", which you can also do.) This is why it's used in HTML; it allows you not just to link to a page, but also to a specific part of that page.

In MEI we recommend using hash fragments for identifying IDs, since there is the idea that you can reference elements in other documents (for example, you can talk about a specific note or measure in some other URI-referenceable notation encoding.) It also ties in with [current research on Linked-Data friendly approaches](https://dh2020.adho.org/wp-content/uploads/2020/07/626_AninteractivemultimediacompaniontoWagnersLohengrinencodingandvisualisingamotivicstudy.html) to music notation encoding, and transformations to LD-friendly formats like RDF and Turtle.

It also helps if you want to tie IDs of musical elements to SVG renderings, and then to CSS styling of the SVG; a consistent ID method means you can carry the ID through the transformations, and then target specific musical encoding elements with standard CSS selectors.

Basically, using the hash fragment method for IDs is not an XML requirement, but it helps tie in to lots of other standards and formats out there.

ahankinson:
You can put newlines in attributes, but most standard XML parsers will remove them since the newline character is not part of the XML spec. (`\n` has no real meaning in XML). It's the same behaviour as in HTML. If you put:

```
<p>
this is a bunch of text
that has a line break in it.
</p>
```

It will render in one line, and the `\n` line ending will be ignored.

notator:
@ahankinson Many thanks! :-)

 **'#' characters**:
All points taken. Enabling a link from an external document seems to me the best reason for using an ID! :-)
So, when the current PR backlog has cleared, I'll submit a PR (for both the Spec and _MNX by Example_) to the effect that MNX recommends the use of a '#' character at the beginning of each reference to an internal ID. Or should MNX _insist_ on such a usage?

**'\n' characters**:
As my proposal stands, its only going to work for one-line texts. That's okay for the usual cases, but more labyrinthine paths though the score (1st goto, 2nd goto etc.) would need some kind of multi-line capability.
I think _that_ would be best realised not in the `<jump>` element (which could have `text=""`), but by having a separate, general purpose, multi-line `<instruction>` direction, that could be used for other kinds of performance instruction as well.
The [current `<instruction>` element](https://w3c.github.io/mnx/specification/common/#the-instruction-element) also only supports one line of text.
We need something more like:
```
<instructionBlock id="ib1" align="right" location="4/4">
    <textLine>text of line1</textLine>
    <textLine>text of line2</textLine>
    <textLine>text of line3</textLine>
    ...
</instructionBlock>
```

**Other things:**
1. I'm not really happy with "jump" as the element name. Its only really a "jump" when its a "goto". So I'd like to refactor it to "branch" or "fork". Has anyone got any better ideas, or would anyone prefer to keep "jump"?
2. I think [this PR](https://github.com/w3c/mnx/pull/190#issue-454338790) should be rejected, but that we should continue to discuss the alternatives here.
3. Having a general purpose `<jump>` or `<branch>` element would help a lot with scores like [this one](http://www.cpdl.org/wiki/images/6/63/The_twelve_days.pdf). Happy Christmas everyone!

notator:
I now think the `<jump>` element should be called a `<switch>`. As in C-like programming languages, it presents a _list_ of possible actions. In the present case, the actions are performed in sequence, but the element is still an n-way switch.

I also have some further ideas about generalising [the current `<instruction>` element in the Draft Spec] to `<text>` and `<textBlock>`s. Those ideas would make the `<target>` element redundant, but they are rather off-topic for this PR, so I'll keep them for later.

ahankinson:
BTW, Entity entries can be used to map custom HTML entities to Unicode codepoints:

```
<!ENTITY segno CDATA "&#x1D10B;"
```

With these types of entries in the top of your XML file, you can use custom Unicode code points for any symbol. This means you could also map them to SMuFL code points for some files, and standard Unicode points for others.

https://www.w3.org/TR/html4/sgml/entities.html

notator:
It occurred to me that we could also add a `<pause>` option to the available actions in the `<switch>` (aka `<jump>`) element:
Its actions list would then contain the following options in any order:
   -  goto somewhere
   -  continue,
   -  pause 
   -  stop.

This would complete the access to the set of controls used by a performing application.
The `<pause>` option would effectively click the application's pause button, and wait for user interaction. It might make presentations a more enjoyable experience (for the presenter) if pauses could be built into the file...

adrianholovaty:
Closing this because it's not relevant anymore:

* MNX uses JSON, and this pull request uses XML
* We've already implemented alternate endings, repeats and segnos

We haven't yet implemented the following:

* Segno
* Support for multiple jumps (e.g., multiple segnos in the same piece)

I'll open separate issues for these two matters so we don't lose track of them.

Issue #193, opened 2020-08-26
Are nested elements within sequences the right approach?


adrianholovaty:
In my work on the [mnxconverter library](https://github.com/w3c/mnxconverter/), I've found that the nested nature of `<tuplet>` and `<beamed>` in MNX has been a bit irritating to deal with.

Figuring out the proper nesting — should `<tuplet>` go inside `<beamed>` or vice versa? — for a given musical situation is non-trivial, at least in the context of converting MusicXML to MNX. [See this commit](https://github.com/w3c/mnxconverter/commit/4df6ebd4048ba7fe3939720be058fa14353fe9f4) for a taste of the complexity.

For example, this music...

<img width="437" alt="Screen Shot 2020-08-26 at 1 47 49 PM" src="https://user-images.githubusercontent.com/180401/91300005-d60b3b80-e7a2-11ea-832a-8f4ef248dae6.png">

...is encoded like so:

```
<beamed>
  <event value="/8">
    <note pitch="C5"/>
  </event>
  <tuplet inner="3/16" outer="2/16">
    <event value="/16">
      <note pitch="D5"/>
    </event>
    <event value="/16">
      <note pitch="E5"/>
    </event>
    <event value="/16">
      <note pitch="D5"/>
    </event>
  </tuplet>
  <event value="/8">
    <note pitch="C5"/>
  </event>
  <event value="/8">
    <note pitch="E5"/>
  </event>
</beamed>
```

And this music...

<img width="242" alt="Screen Shot 2020-08-26 at 1 48 19 PM" src="https://user-images.githubusercontent.com/180401/91300020-db688600-e7a2-11ea-968a-7053a57707c5.png">

...is encoded like so:

```
<tuplet inner="3/8" outer="2/8">
  <event value="/8">
    <rest/>
  </event>
  <beamed>
    <event value="/8">
      <note pitch="C5"/>
    </event>
    <event value="/8">
      <note pitch="E5"/>
    </event>
  </beamed>
</tuplet>
```

In the first example, the outer element is `<beamed>`. In the second example, the outer element is `<tuplet>`. The nesting differs depending on the musical situation.

This nested nature means that there's developer friction when doing the following with MNX:

* When reading MNX, get a simple list of the notes/rests in a bar (friction: you need to dive into the nested tuplets/beams; the notes/rests aren't just at a uniform top level).
[...]

PR #194, opened 2020-09-28
Remove <beamed> from spec and by-example

adrianholovaty:
This pull requests removes `<beamed>` as a structural element — both from the spec and from "MNX By Example." See #193.

For now, it hasn't been replaced by any new/different approach. Eventually we'll need to determine a different way of specifying beaming, in a more presentational (less structural) way.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/pull/194.html" title="Last updated on Sep 28, 2020, 6:43 PM UTC (3ff5af0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/194/f30a4e2...3ff5af0.html" title="Last updated on Sep 28, 2020, 6:43 PM UTC (3ff5af0)">Diff</a>


adrianholovaty:
@dspreadbury @mdgood Could you review when you get a chance?

specification/common/index.bs
```diff
@@ -3161,6 +3003,8 @@ notes in a swung triplet rhythm instead:
 ```
 </div>
 
+TODO: `<beamed>` no longer exists, so the previous example is invalid.
```
mdgood, line 3006: Could we just remove the previous example and then reword the introductory text to the next example? I think that should still cover the use of the performance-event while cleanly removing the beamed element from the spec.

mdgood:
Looks good - just one suggestion regarding the TODO.

specification/common/index.bs
adrianholovaty: @mdgood Thanks — I've just updated the pull request to handle that. Look OK?

mdgood (approved):

PR #201, opened 2020-10-05
Complete the renaming of MNX-Common to MNX (Part 1)

notator:
This Pull Request is the first of a pair of PRs, that together complete the renaming of MNX-Common to MNX.
**Part 1** (this PR) consists of a revision of the [MNX Draft Specification](https://w3c.github.io/mnx/specification/common/#the-score-element) in which:  
   1. The current `<mnx>` definition (the MNX-Container) has been removed.
   2. The `<collection>` element has been removed (it was required only by the MNX-Container).
   3. The `<mnx-common>` element has been renamed `<mnx>`.
   4. The definition of `<score>` has been corrected and a note has been added.
This element's context and content are currently under discussion. See the [Co-Chair Meeting Minutes: July 7, 2020](https://www.w3.org/community/music-notation/2020/07/07/co-chair-meeting-minutes-july-7-2020). 
[Edit 07.10.20: The `<score>` element has now been _deleted_. It is not part of the current `mnx-common` element, so does not belong in this revision.]
   5. Other texts have been updated only to ensure that the document is consistent.
The intention is to preserve the content of the current document as exactly as possible while renaming `<mnx-common>` as `<mnx>`.

**Part 2** (a separate PR) adds a new **MNX-Container Draft Specification** to this repository.
In the new document, the `<mnx>` element that was removed by this PR, is replaced by an `<mnx-container>` element.

***
<a href="https://pr-preview.s3.amazonaws.com/notator/mnx/pull/201.html" title="Last updated on Oct 24, 2020, 12:32 PM UTC (88b4ad0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/mnx/201/da4d333...notator:88b4ad0.html" title="Last updated on Oct 24, 2020, 12:32 PM UTC (88b4ad0)">Diff</a>

### DIFF
--- a/by-example/index.html
+++ b/by-example/index.html
@@ -100,30 +100,26 @@
             <h3>MNX</h3>
             <div class="markupcode">
             <script type="application/xml" class="markupxml"><mnx>
-    <score>
-        <mnx-common>
-            <global>
-                <measure>
-                    <directions>
-                        <time signature="4/4"/>
-                    </directions>
-                </measure>
-            </global>
-            <part>
-                <part-name>Music</part-name>
-                <measure barline="regular">
-                    <directions>
-                        <clef sign="G" line="2"/>
-                    </directions>
-                    <sequence>
-                        <event value="/1">
-                            <note pitch="C4"/>
-                        </event>
-                    </sequence>
-                </measure>
-            </part>
-        </mnx-common>
-    </score>
+    <global>
+        <measure>
+            <directions>
+                <time signature="4/4"/>
+            </directions>
+        </measure>
+    </global>
+    <part>
+        <part-name>Music</part-name>
+        <measure barline="regular">
+            <directions>
+                <clef sign="G" line="2"/>
+            </directions>
+            <sequence>
+                <event value="/1">
+                    <note pitch="C4"/>
+                </event>
+            </sequence>
+        </measure>
+    </part>
 </mnx></script>
             </div>
         </div>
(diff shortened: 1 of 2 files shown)


notator:
The **score** element is not part of the current mnx (=mnx-common) element, so should not be included here.

adrianholovaty:
Thanks for taking on this work, @notator.

Could you post a quick comment to this thread, with the ["Hello World" MNX example](https://w3c.github.io/mnx/by-example/#helloworld) encoded with your changes? This would help me understand the scope of the changes you're proposing here. Sorry, reading spec changes is a bit too indirect for my brain to intelligently parse. :)

After you've done that and I confirm it looks OK, here's some feedback on the pull request itself:
* I'd like to avoid introducing a whole other spec (as you've done in part two of your pull request, #202). Can you find a way to include this in the MNX spec directly?
* Could you also update the files in MNX By Example? This simply involves changing the expected MNX markup in every example. Happy to answer any questions about the markup/formatting there.

notator:
You're welcome! :-)

The effect of this pull request is to delete everything outside the original `<mnx-common>` element (so that it becomes the first element in the file) and then rename it as `<mnx>`.
```
<mnx>
    <score>
        <mnx-common>
            ...
        </mnx-common>
    </score>
</mnx>
```
becomes
```
<mnx>
    ...
</mnx>
```
The ["Hello World" MNX example](https://w3c.github.io/mnx/by-example/#helloworld) becomes:
```
<mnx>
    <global>
        <measure>
            <directions>
                <time signature="4/4"/>
            </directions>
        </measure>
    </global>
    <part>
        <part-name>Music</part-name>
        <measure barline="regular">
            <directions>
                <clef sign="G" line="2"/>
            </directions>
            <sequence>
                <event value="/1">
                    <note pitch="C4"/>
                </event>
            </sequence>
        </measure>
    </part>
</mnx>
```

> Can you find a way to include [the MNX-Container spec] in the MNX spec directly?

I'm afraid it would make no sense to include the MNX-Container Draft Spec in the MNX Draft Spec:
These are actually two different specs, so they belong in different files (just as MNX-Generic has its own file).
As defined in the MNX-Container Draft Spec, MNX-Container files can contain `<mnx>` scores. But it makes no sense to say (in the MNX Draft Spec) that MNX files can _be contained in_ MNX-Container files. MNX files can _be contained in_ lots of different contexts (e.g. .zip files), but the MNX Draft Spec can have nothing to say  about things like that...
If the MNX-Container spec were to be included in the same file with the MNX spec, then we'd have to change the title to "MNX-Container and MNX Draft Specs" - which would be awful.
I just think things stay a lot simpler if we keep the MNX-Container, MNX and MNX-Generic in three separate files.

>Could you also update the files in MNX By Example?

Will do. Might take a day or two...

notator:
Done! :-)

adrianholovaty:
Thanks @notator — this is now merged!

PR #202, opened 2020-10-05
Complete the renaming of MNX-Common to MNX (Part 2)

notator:
This PR adds the **MNX-Container Draft Specification** to the repository.
This is a new document, that replaces the MNX-Container definition (the `<mnx>` element) removed from the MNX Draft Spec by Pull Request #201. 

Caveat: This file may be redundant.
The [Group Charter](https://www.w3.org/community/music-notation/wiki/Group_Charter) describes the MNX container format as follows (my italics):
> This format describes and relates an arbitrary number of components and documents which collectively describe _a notated musical work_ as a whole.

Note that the MNX container format seems originally to have been intended to accommodate different scores for _a single musical work_, and that Example 6 (=Example 2 in this new document) illustrates how to include _parts_ with a full score.
Including separate scores for the parts and full-score is a maintenance problem (for which we now have a separate solution), and  other information about "the musical work as a whole" can be included in a .zip file. So I think this container format is redundant, and that the co-chair could simply leave it out when [re-drafting the Charter](https://www.w3.org/community/music-notation/2020/09/01/co-chair-meeting-minutes-september-1-2020/)

The document was created as follows:
1. A new folder "mnx/specification/**container**" was created next to the existing "mnx/specification/**common**" folder
2. A copy of the current [MNX Draft Specification](https://w3c.github.io/mnx/specification/common/index.bs) was made in the new folder at "mnx/specification/**container**/index.bs"
3. In "mnx/specification/**container**/index.bs"
   1. The title was changed.
   2. The definition of `<mnx-common>` was removed
   3. The `<mnx>` element was renamed `<mnx-container>`
   4. The file was simplified as far as possible, without changing the meaning of the `<mnx-container>` definition. Where possible, duplicate MNX definitions are avoided by providing links to the definitions in the main [MNX Draft Specification](https://w3c.github.io/mnx/specification/common/index.bs).

---

Issues:
1. The [MNX-Generic Draft Specification](https://w3c.github.io/mnx/specification/generic/) may need to be updated: Note that I have re-included `<mnx-generic>` as a score type, as it was before MNX-Generic was moved out of the original MNX Draft Specification. That's just for the record, since I think that MNX-Generic should also be omitted from the next version of the Charter. If MNX-Generic is _**not**_ going to be removed, then it needs to be revised, especially to remove duplicate definitions.
2. **Also in the main [MNX Draft Specification](https://w3c.github.io/mnx/specification/common/)**: The definition of "Metadata content" belongs in the section about "Infrastructure", not in "Document structure". That's because the definition depends on external definitions (i.e. "infrastructure"). I've already done that here, to show how I think the main spec should look.


notator:
The **score** element contained by the mnx-container _contains_ an mnx (=mnx-common), but is not _contained_by_ mnx elements. So it is not the same element, and should not be linked to a definition inside mnx.

adrianholovaty:
(I've posted a comment to #201 referencing this pull request.)

notator:
Following the [Co-chair Meeting Minutes:  October 13, 2020](https://www.w3.org/community/music-notation/2020/10/13/co-chair-meeting-minutes-october-13-2020/), its not entirely clear to me what the co-chair wants to do with this PR.
If we are not going to be changing the charter after all for the time being, I'd suggest that the MNX-Container Draft Spec should be created and then forgotten. It needn't get in the way of the MNX debate. As with MNX-Generic, it can be deleted at the proper time.

adrianholovaty:
This additional spec is something we might want to tackle in the future, but I don't think it's necessary at this time. Marking as closed, but the code/patch will always be here when we're ready to revisit.

PR #232, opened 2021-04-15
Visual design

edwardoriordan:
I've got a start on adding some basic visual design to this now.

* I'm a bit unsure where to go next. I assume the content might change (for example, so the sidebar will do some of the heavy lifting of navigation so the index page would be more of a document page with intro text, etc). I could optimise the design for the current content but I'm a bit worried it might not work with future content
* I added a little hamburger menu primarily because it balances out the header section a bit! It would be handy to show/hide the sidebar on mobile. I didn't add an JS to actually make it work. Any thoughts on if we should bother with it? We could put the W3C logo in a different place, perhaps not in the header. Like on the initial intro page or maybe even in the sidebar.


mdgood:
Thanks @edwardoriordan!

Regarding the header section and the logo, what other W3C Community Groups seem to do is to keep the "W3C Community Group Final Report" logo at the top of the sidebar, and only include the W3C logo on the title page of the report. We tried to mimic that for the MusicXML 3.1 report at 

https://www.w3.org/2017/12/musicxml31/

You can find another example at

https://www.w3.org/2019/08/did-20190828/

In our SMuFL report the header lets you change font size and display settings, and also lets you hide the sidebar. It also makes the previous page / next page navigation areas more discoverable. I think GitBook is providing this automatically. You can see what it's doing at 

https://www.w3.org/2021/03/smufl14/

Perhaps we could use that type of header, or possibly remove the header altogether?

In terms of logos, the MusicXML approach of the logo in the sidebar is closer to W3C Community Group recommended style. SMuFL puts it in a different place because placing it in the sidebar would have involved much fighting with GitBook.

I can email you the current MusicXML database so you can see what the design will look like in that context. I'm not quite ready to turn that into a pull request in the MusicXML repo yet, but hopefully soon.

For our drafts we will want to avoid using the final community report logo. I haven't yet found a draft report logo, but perhaps we could create a dummy or draft logo of the same size that we swap out at the end.

mdgood:
Found a draft logo: https://maps4html.org/HTML-Map-Element-UseCases-Requirements/

mdgood (approved):
This looks good to me. Should we merge now and do further development in another pull request? Whatever works best for you is fine with me.

adrianholovaty:
All looks good — merged now! Yes, I agree that any further work can be done on separate branches / pull requests. Thanks @edwardoriordan

PR #234, opened 2021-04-30
Changes from MusicXML documentation

mdgood:
Hopefully these work OK for MNX too.


adrianholovaty:
I've just made a series of commits to integrate all of these changes (using separate commits for separate concepts to make things more atomic). They're all in the MNX repo now. Thanks @mdgood!

PR #236, opened 2021-05-13
Docgenerator: Example style change for MusicXML

mdgood:


docgenerator/spec/templates/example_detail.html
```diff
@@ -19,7 +19,7 @@
 <h1>{{ example.name }}</h1>
 
 {% if example.image_url %}
-<p width="100%"><img src="{% relative_url_string example.image_url %}" width="100%"></p>
+<p><img src="{% relative_url_string example.image_url %}"/></p>
```
adrianholovaty, line 22: We're not using XHTML for this site, so the slash at the end of  isn't necessary (to be consistent with the rest of the templates).

adrianholovaty:
@mdgood Thanks — all looks good, except would you mind removing those slashes (per my review comment)?

mdgood:
Done. Old XML habits die hard. I've got the slash at the end of all the hard-coded `` elements within the description field.

adrianholovaty:
Fixed this in ce713c49a. Thanks!

PR #241, opened 2021-05-21
Fix content formatting on element detail page

mdgood:


adrianholovaty:
I considered making this exact change a few days ago, but unfortunately it results in inconsistent layout. It works perfectly if there's only one child relationship, but in other cases it inserts undesired space, like after "In this order" on `<note>` —

Screen Shot 2021-05-21 at 7 19 21 PM

I've tried a different approach in 2c3a317e — which I assume addresses the issue you were trying to fix here. Could you give it a shot and let me know whether that's solved the layout issue?

mdgood:
Yes, the change in 2c3a317 solves the layout issue just fine. Thanks!

Issue #278, opened 2021-12-13
Multi-line text blocks


notator:
Multi-line text blocks have been discussed before, but we don't seem to have reached any conclusion. I think we need them, and that they would be of general use.

In #277, I used a temporary strategy that involved defining the following two elements:
   - `<string>`:
      - Attribute `text` (a [String](https://w3c.github.io/mnx/docs/mnx-reference/data-types/string/)): defines a string of text that will be rendered in a single line.
      - Content: none
   - `<text-block>`:
      - Attibutes: none
      -  Content: one or more `<string>` elements that define the lines of a multi-line text block.

Can this be improved? Is there a better way to do it?

joeberkovitz:
I agree that we need them. I foresee style properties governing their font size, dimensions, borders and padding/margins in a natural way. Ultimately there should be a way of including these elements directly in a `<page>` so that text blocks can be freely interpolated with systems, e.g.
```xml
   <score>
       <page>
        <system .../>
        <text-block .../>
        <system .../>
      </page>
   </score>
```

clnoel:
I like this in general. We should say that any element that has [String](https://w3c.github.io/mnx/docs/mnx-reference/data-types/string/) _content_ (not an attribute) should be changed to take a `<text-block>` instead. Right now, this is directions->expression, directions->instruction, and part->part-abbreviation. (Hmm... do we even still need that last one?)

Caveats:
1. The `text-block` needs a `justification` attribute that governs how the strings in the block justify _relative to each other_. (left|right|center)
2. This is not a new issue here, but I want to make it again, since I can't find it anywhere: This String type assumes the same font for the entirety of the string. We might need to allow for `<em>` and `<strong>`, where the fonts to be used by those tags get defined by the style properties. And allow for SmufL code-points using something like `&segno;`, which I remember being proposed somewhere but can't find right now. This would involve changes/clarification to the String type.

samuelbradshaw:
Instead of specifying a list of HTML tags like `<em>`, `<strong>`, etc. that can be borrowed and used in text blocks, what are your thoughts about supporting an `<html-block>` that can take HTML directly?

For example, here are two examples of credits from a hymnal. The formatting includes:
- Specific font family and font size that isn't necessarily the same as the lyrics
- Paragraphs with a hanging indent (`<p>` with a class)
- Italicized text (`<i>Text:</i>`)
- Emphasized text (`<em>works</em>`)
- Small-caps text (`<span>` with a class)
[...]

Issue #277, opened 2021-12-13
How to code a system's left-margin labels


notator:
This proposal
- takes #276 into account, and should also help to resolve #269.
Note that the `<group-layout>` element is no longer recursive here, but the `<group-label>` element _is_. See the _Thoughts about Recursion_ below.
- separates left-margin labels (the labels and brackets to the left of a system), from the `<part>`-related _content_ of the system's staves. The  `<part>` element can't know how it is going to be combined on a staff with other `<part>`s, so can't know exactly how it is going to be labelled.
- encloses the left-margin labels in a `<labels>` element, inside `<system-layout>`, following the `<staff-layout>` and `<group-layout>` definitions.
- uses system staff indices, in a way that is similar to `<part>`s local use of [staff index](https://w3c.github.io/mnx/docs/mnx-reference/data-types/staff-index/). 
- makes `<part>`'s `name` and `short-name` attributes redundant. They should be deleted. 
- makes the existing `label`, `labelref` and `symbol` attributes on `<group-layout>` and `<staff-layout>` redundant. These attributes should be deleted.
- means that `<group-layout>` no longer needs to be recursive.
`<group-layout>`'s  `<group-layout>` content should therefore be deleted.

As a reminder of how to code a grand staff, see @clnoel's https://github.com/w3c/mnx/issues/185#issuecomment-974201595 (point 5).

<details>
<summary>Thoughts about Recursion</summary>

Simple recursion is possible in XML schemas (See
https://stackoverflow.com/questions/148988/recursion-in-an-xml-schema).
If the following two-level recursion is a problem
```
<group-label>
    <group-symbol>
        <group-label>
```
we could simply create two new elements (`<group-label2>` and `<group-symbol2>`), and limit the nesting to two levels:
```
<group-label2>
    <group-symbol2>
        <group-label>
            <group-symbol>
```
This would only affect Example 4 below.

If thought necessary, three levels could also be implemented in this way.

</details>

---

<details>
<summary>Definitions of Proposed New Elements</summary>

The code in the rest of this proposal assumes the following new elements:
- two low-level helper elements for creating multi-line text blocks.
(We've discussed this before, but don't seem to have come to any conclusion. I've adopted this solution here as a temporary measure. These details could be changed later.):
   - `<string>`:
      - Attribute `text` (a [String](https://w3c.github.io/mnx/docs/mnx-reference/data-types/string/)): defines a string of text that will be rendered in a single line.
      - Content: none
   - `<text-block>`:
[...]

PR #281, opened 2022-01-30
More sophisticated layouts (from issue #185)

adrianholovaty:
This pull request brings the agreed-upon changes from #185 (specifically, from [this summary comment](https://github.com/w3c/mnx/issues/185#issuecomment-987019139)) into the MNX spec. This introduces more sophisticated methods of mixing and matching notation data within layouts.

The updated docs (as of this pull request) are visible here: https://cdn.githubraw.com/w3c/mnx/issue-185/docs/

The specific changes are:

* Added [&lt;voice-layout&gt;](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/elements/voice-layout/), which lives within &lt;staff-layout&gt;. This is a way to say "In this staff, use the notation data from parts X and Y; use upstems for all of part X's notes and use downstems for all of part Y's notes."
* Added [&lt;system-layout-change&gt;](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/elements/system-layout-change/). This is a way for a system layout to change mid-measure.
* Gave [&lt;part-layout&gt;](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/elements/part-layout/) two new attributes, `stem` and `voice`. Note that I opted for `voice` instead of `part-voice`, to be consistent with the existing `staff`.
* Added the example document [Multiple layouts](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/examples/multiple-layouts/).
* Added the example document [Orchestral layout](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/examples/orchestral-layout/).
* Added the example document [Organ layout](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/examples/organ-layout/).
* Updated the existing example document [System layouts](https://cdn.githubraw.com/w3c/mnx/issue-185/docs/mnx-reference/examples/system-layouts/) to use &lt;voice-layout&gt;.

Special thanks to @clnoel for putting together the example images and markup! Any mistakes are my fault.

For everybody reviewing this, I _highly_ recommend looking at those four example documents to gain some intuition on how this stuff all works and what the benefit is. Speaking personally, I found it hard to understand the new elements/attributes in the long-winded discussion of #185, and seeing specific images made it much clearer.

One final note. The original summary suggested `<staff-layout>` needs a `part-staff` attribute, but I didn't see the need to add that in this pull request. Wouldn't this already be handled by the child `<part-layout>`'s `staff` attribute? Let me know if I'm missing something.


ahankinson:
You might consider changing the examples to read "Upper voices" and "Lower voices" instead of "Women" and "Men" since the voice parts are not exclusive to a given gender.

bhamblok (changes requested):
The HTML-element `<nobr>` is a non-standard and deprecated element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr
We could use `<span style="white-space: nowrap;">` instead.

clnoel:
1. Hmm.. the pre-existing `staff` attribute on `<part-layout>` was supposed to be removed as a part of this, in favor of moving it to the `<staff-layout>` (I forgot to list that in my summary post). In retrospect, however, I prefer having it on the `<part-layout>`. I think it makes more sense. I'd appreciate other's viewpoints on that, though.
2. I dislike the change from `part-voice` to `voice`. I prefer to change `staff` to `part-staff` instead. This is a very clear indicator, to me, that we are filtering the part itself, instead of saying, for example, that the `<part-layout>` is defining a voice id, or is on the second staff down. 

Suggested changes:
* Regardless of the decisions for the two points above, the examples should match the definitions and right now they do not. `part-voice` is used, even though `voice` is defined in the definitions, and `part-staff` is used on `<staff-layout>` even though it is not defined on that element.
* I'm not sure if it was you or me, @adrianholovaty , but I think that in the "Organ Layout" the "organ3StaveSplitOber" layout should use `part-staff=2` instead of `part-voice="Hauptwerk"` and `part-staff=3` instead of `part-voice="Pedal"`, since that example was supposed to demonstrate the use of `part-staff`.

notator:
I agree with @clnoel  on all points.

Also, I think we need to resolve #278 before committing this PR.
How should the label for a staff containing two parts be encoded?
twoPartsOneStaff

clnoel:
@notator, while the labels are included in the examples, how we do the labels is not a focus, which is why I did not attempt to recognize the line breaks in the labels of the Orchestral Layout example at all. We can come back later and alter that example once we've hashed it out. I've finally responded to #277, which actually has a potential for greater impact on the new examples, if you want to go examine that. But I don't think that we need to hold up this PR for either unresolved issue. We'll just make a new PR with the changes in it.

adrianholovaty:
@ahankinson Thanks — that's now fixed in the example ("SA" and "TB").

@clnoel Thanks — I agree `part-voice` and `part-staff` are clearer. I've just updated the pull request accordingly, and I've fixed the incorrect bits in the example docs. Could you give it a final readthrough and make sure everything looks good? I believe this is ready to merge but will wait for your OK.

clnoel:
@adrianholovaty I think you either misunderstood my suggestion for the Organ example, or made a typo.

In the "organ3StaveSplitOber" layout, the following line:
`<part-layout part="organ" part-staff="2" stem="down"/>`
should instead be
`<part-layout part="organ" part-voice="Oberwerk" stem="down"/>`

and

`<part-layout part="organ" part-voice="Hauptwerk"/>`
should instead be:
`<part-layout part="organ" part-staff="2"/>`


Other than that, I think it looks really good!

--Christina

adrianholovaty:
@clnoel Thanks for the catch! That's now fixed. I'll go ahead and merge this now.

Issue #284, opened 2022-07-04
Add support for multi-bar rests


adrianholovaty:
Here are a few proposals for how MNX can handle multi-bar rests. As you'll see, there are some open questions around where this data should live and how much flexibility MNX should provide.

We'll use the following music in each example. It consists of a full score, plus two individual parts:
![mnx_score_parts](https://user-images.githubusercontent.com/180401/177207561-f077ac71-f382-4ae0-b356-e6f0651f1c16.png)

Some things to note, before I get into proposals:

* The two parts each have multi-bar rests. Part A has one, and Part B has two.
* Multi-bar rests are only rendered in the individual parts, not in the full score. The full score simply uses "normal" whole rests for those bars.
* To be precise about terminology: I'm using the word "part" to mean the rendered music notation as typically distributed to musicians playing a specific instrument and in contrast with the "full score." I'm using the syntax-highlighted term `<part>` when referring to the contents of the MNX `<part>` element (which contains the semantic notation data for a given instrument and is in turn used when generating the full score and parts).
* In chatting about this with my fellow co-chairs, we agreed that a rules-based approach is off the table. It's tempting to encode something like "In this part, find any contiguous sections of at least two empty measures, and automatically convert those into multi-bar rests" — but we need to handle multiple multi-bar rests next to one another (deliberately split into multiple multi-bar rests to communicate musical structure or time/key signature changes, etc). So we need to be explicit about the multi-bar rest locations rather than encoding implicit rules.

Here is the basic MNX document for this notation data, excluding any multi-bar rest information:

```xml
<mnx>
  <global>
    <measure-global>
      <directions-global>
        <time signature="4/4"/>
      </directions-global>
    </measure-global>
    <measure-global/>
    <measure-global/>
    <measure-global/>
    <measure-global/>
    <measure-global/>
    <measure-global/>
  </global>
  <part id="PartA">
    <measure>
      <directions-part>
        <clef sign="G" line="2"/>
      </directions-part>
      <sequence>
        <event value="/4">
          <note pitch="C5"/>
        </event>
        <event value="/4">
          <note pitch="E5"/>
        </event>
        <event value="/2">
          <note pitch="G5"/>
        </event>
      </sequence>
    </measure>
    <measure>
      <sequence>
        <event value="/4">
          <note pitch="C5"/>
        </event>
        <event value="/4">
          <note pitch="E5"/>
        </event>
        <event value="/2">
          <note pitch="G5"/>
        </event>
      </sequence>
[...]

PR #285, opened 2022-08-16
Add multimeasure rests to spec, with example

adrianholovaty:
This pull request adds multimeasure rests to the spec, along with an example document. This is based on the agreed-upon approach from #284.

The updated docs (as of this pull request) are visible here: https://cdn.githubraw.com/w3c/mnx/multimeasure-rests/docs/

Here's the example document: https://cdn.githubraw.com/w3c/mnx/multimeasure-rests/docs/mnx-reference/examples/multimeasure-rests/

@cecilios and @clnoel — mind reviewing to make sure everything matches our discussion?


clnoel:
This looks like what we agreed, although I might have added the `page` and `system` tags into the part scores in order to clarify the sibling relationship.

However, I am now throwing a wrench in the works. We either need to decide to support the old-style rest format, or decide to put it off until later. We shouldn't forget about it. As recently as last year people were trying to add this to [MuseScore](https://musescore.org/en/node/44846). 
image
--Christina

bhamblok (approved):

adrianholovaty:
@clnoel Thanks! I've just created issue #286 to collect discussion/ideas for the old-style format. I don't think the lack of that should block this particular pull request, but I do want to make sure we don't lose track of it.

PR #302, opened 2023-06-08
Retweak top of homepage

mscuthbert:
A tweaking of the top of the homepage, letting people know (1) how MNX is different from other notation formats, and (2) that it's not the place to start for new users until it's been released.

I plan to work the rest of the page too, but this seems like a good start to keep up the JSON excitement.

Glad to accept tweaks or proposed changes directly on this PR.


adrianholovaty:
Thanks — I've just made some small copy tweaks and made sure the changes were made in _two_ places (the template and the generated HTML). See 2dd99c9d for the commit.

Issue #307, opened 2023-08-31
JSON: Example Accidentals, boolean property values must not be quoted


paul-bayleaf:
The MNX example Accidentals has quoted the boolean property value "true".
Lines 69, 125, 167.

Suggested fix: remove quotes from boolean values.

I have written a JSON schema for MNX, is this of interest?  If so I can prepare a pull request.

mscuthbert:
Yes! It! Is!

adrianholovaty:
@paul-bayleaf Thanks for another good catch! Yes, please do open a pull request with a JSON Schema. My plan is to try to automatically generate such a schema from our documentation, but it would be very useful to get another person's perspective on it (and it would definitely come in handy in the interim). Much appreciated.

williamclocksin:
[...]

Issue #306, opened 2023-08-23
JSON: Discrepancies between the MNX examples and the Objects' documentation.


paul-bayleaf:
I don't know if it is too early to raise issues with JSON MNX, but there are some discrepancies between the MNX examples and the Objects' documentation.  These are all minor and easily fixed.

Examples https://w3c.github.io/mnx/docs/mnx-reference/examples/
Objects https://w3c.github.io/mnx/docs/mnx-reference/objects/

**1) Event Object property "type" is required but is missing from the following MNX examples:**

Beams (with inner grace notes) 
lines 57, 73, 87, 101, 144

Grace notes (beamed) 
lines 70, 84, 117, 131, 145, 178, 192, 206, 220

Multiple layouts
lines 205, 218, 231, 267, 280, 328, 341, 354, 390, 403, 451, 464, 477, 513, 526, 539, 587, 600, 613, 649, 662

Tempo markings
lines 52, 65, 78, 114, 127, 140

Suggested fix: add missing property "type": "event" to the "event" objects on the lines above.

**2) String property values must be quoted, but the quotes are missing from:**

Jumps (Dal Segno) 
Unquoted string property value on "location" lines 12, 20

Jumps (D.S. al Fine)
Unquoted string property value on "location" lines 12, 17, 24

Suggested fix: add quotes to the above string property values.

**3) Numeric property values must not be quoted, but they are here:**

Key signatures 
Quoted numeric property value "fifths" line 16

Octave shifts (8va) 
Quoted numeric property value "value" line 60

Suggested fix: remove quotes from the above numeric property values

**4) Global Data Object property "measures" is required, but is missing from:**

Orchestral layout 
Empty object "global" line 1

System layouts 
Empty object "global" line 1

Suggested fix add property "measures": [] to the above global objects.

**5) Stave Source Object does not have property"staff" but is used as follows:**

Orchestral layout
"staff" on lines 48, 45

Organ layout
"staff" on lines  lines 27, 36, 48, 91

System layouts
"staff" on lines  lines 70, 79, 156, 165

Suggested fix: in the documentation for Stave Source Object add property "staff".

**6) Stave Source Object does not have property "labelref" but is used as follows:**

System layouts
"labelref" on lines 19, 24, 48, 53, 134, 139

Suggested fix: 

I note that "labelref" is a property of the enclosing Stave Object, should this have been used instead?
Otherwise in the documentation for Stave Source Object add property "labelref".

[...]

PR #308, opened 2023-08-31
Implement JSON schema

paul-bayleaf:
Implements a JSON schema for MNX.  Tested against all the examples 
https://w3c.github.io/mnx/docs/mnx-reference/examples/

All verify with the exception of Accidentals (issue #307) and Orchestral layout and Systems layouts which fail by design (issue #306).   Tested using jsonschema v3.2.0 under openSUSE Linux v15.5.

All objects have "additionalProperties": false, this is to spot mistyped optional properties, we might want to reconsider this for production.


webern:
Awesome!

I think there needs to be some .github actions workflow added that checks examples against the schema. Doesn't have to be with this PR but the schema is certain to be immediately out-of-date without an automated check.

What did you use to test the schema against examples? Paste your script here?

paul-bayleaf:
Thank you. I agree using a schema to automatically test the examples, and equally using the examples to test the schema, would be very useful.   Currently, as far as I can see, the examples only exist in this repo as html, so in order to test them I had to manually copy each example from https://w3c.github.io/mnx/docs/mnx-reference/examples/ and paste into a corresponding .mnx text file.  To automate testing we need to either generate .mnx files from the html, or perhaps better, generate the html from the .mnx files.

I used jsonschema under Linux openSUSE 15.5 to test my schema against the examples.  (Other validators are available).

My script is nothing special, it simply invokes jsonschema for each .mnx file with my schema.

jsonschema -i file.mnx mnx_schema.json

webern:
Cool, that would be pretty easy to script in github actions provided the examples could be separated from the HTML in the repo.

paul-bayleaf:
I've updated the schema to reflect pull request #312 .   I'll aim to keep this schema up to date with future developments.

adrianholovaty:
@paul-bayleaf Thanks Paul! This is still on my to-do list to review, with the eventual goal of automatically generating the schema from our documentation system (using your work as a guide).

paul-bayleaf:
I've updated the schema to reflect commit 98b1a43, "Added way to encode inner-bar clef changes".  Example [Clef changes](https://w3c.github.io/mnx/docs/mnx-reference/examples/clef-changes/) validates.

paul-bayleaf:
I've updated the schema to reflect commit c5e1db9 , "Added staff position type; changed clef to use it".

adrianholovaty:
Closing this now, since we've added an automatically generated JSON Schema. Big thanks again to @paul-bayleaf for helping make this happen!

Issue #96, opened 2018-04-09
clef changes within beamed groups


jsutdolph:
![screen shot 2018-04-09 at 11 42 08](https://user-images.githubusercontent.com/13803204/38493733-13c37bb0-3beb-11e8-9d90-893d569f72ab.png)
I have read the spec on beamed groups and I don't see how clef changes within groups are supported. Probably just a misunderstanding

clnoel:
The `<clef>` element, like all directions, has a location attribute. In the above work,  the starting clef is in the global, with location 0. The next clef would have location 1/8. What's not clear in the spec is whether the second clef should be in the global or be part of the sequence.

clnoel:
@joeberkovitz A sequence specific clef might actually make sense, like in this example that was pulled from the MusicXML conversation on almost this same topic from two years ago:

![image](https://user-images.githubusercontent.com/6453511/38585009-ec7fd1ea-3cdd-11e8-9fa2-21ca43c2c7e1.png)

Not exactly the current topic, but definitely related!

joeberkovitz:
Oh yeah, forgot about that one! Although it's probably outside the well-behaved boundary we'd still need to decide if a sequence-specific clef is the way that things like the above are encoded.

adrianholovaty:
Closing this in favor of #313.

Issue #45, opened 2017-12-20
Ordering of grace notes relative to other notation


joeberkovitz:
CWMNX needs to allow scores to encode the notational ordering of grace notes relative to other objects. Such an ordering is well defined for non-grace notes, but there are particular problems with grace notes.

As a somewhat artificial scenario, consider this case:
- A staff in some measure contains two separate voices
- Voice 1 starts with a run of 4 grace notes (acciacatura), followed by a normal note
- Voice 2 starts with a run of 4 grace notes, followed by a normal note
- A change from F clef to G clef occurs just after grace note 1 of Voice 1, and just after grace note 2 of Voice 2. In other words, the F clef applies to the first grace note of Voice 1, and to the first two grace notes of Voice 2.
- Only a single G clef appears: there's no switching back and forth between F and G.

Questions:
-  is the clef change specified at measure level (once), or at voice level (twice)?
- if at measure level, how is its effect on the grace notes known?

clnoel:
Here is a _real example_ of a piece that came into us (Musicnotes) recently, that completely and utterly failed somewhere in the Sibelius MusicXML export and our MusicXML import. We ended up with the clef change before all the grace notes, and the octava applying only to the regular note the grace notes are attached to. I would like to make sure that MNX has a well-defined way to do this.

![image](https://user-images.githubusercontent.com/6453511/152571539-c4db4b7e-d89e-47c7-8e68-9b80539fa0de.png)


As I note that the `<grace>` element currently does not allow directions at all, even though it looks like it's supposed to based on its explanatory text, I thought I would add this example into the mix.

mdgood:
@clnoel This is likely a deficiency in the MusicXML export rather than in MusicXML. Directions can be placed before and after grace notes just like they can with regular notes. The problem for our plug-in is likely that all these events are represented to the plug-in at the same musical time, so you can't tell where the `<octave-shift>` should appear in the MusicXML output. I'm not sure why native export would have the same issue, but perhaps it's related.

If you email me the original Sibelius file I can take a look. I'm also not sure why the clef change would be a problem in this example. We just released a new version of the Dolet for Sibelius plug-in, so maybe there are things to fix in a maintenance update. Perhaps there is even some additional information we can use for the `<octave-shift>` that we overlooked before.

clnoel:
@mdgood Thanks! I'll be communicating with you about that outside this issue. I still feel like making sure that that example works in a _defined way_ for MNX is important.

[...]

PR #312, opened 2023-09-17
Change part-measure objects to use separate arrays for clefs and sequences

adrianholovaty:
In my work to update the [mnxconverter](https://github.com/w3c/mnxconverter) to output JSON instead of XML, I realized the current approach of encoding measure contents into a heterogeneous list (containing clefs and sequences) has started to feel awkward. I brought this up in the latest co-chairs meeting ([see here](https://www.w3.org/community/music-notation/2023/09/14/co-chair-meeting-minutes-september-14-2023/)), and the three co-chairs agreed it's worth simplifying the encoding.

This pull request addresses that. The specific changes are:

* In the [part-measure object](https://cdn.githubraw.com/w3c/mnx/sequence-simplification/docs/mnx-reference/objects/part-measure/): instead of a `"contents"` key that contains clefs and sequences, we now use two separate keys, `"clefs"` and `"sequences"`.
* That `"clefs"` key is a list of [positioned clef](https://cdn.githubraw.com/w3c/mnx/sequence-simplification/docs/mnx-reference/objects/positioned-clef/) objects, each of which has a clef and position. The position is currently encoded as a two-element array, where `[1, 4]` would be a quarter note's duration into the measure.
* In a positioned clef, the position is optional. If not provided, the position should be interpreted as zero (i.e., the start of the measure). This is a small nicety to make the common case simpler to encode.
* Clef objects no longer require a `"type": "clef"`, and sequence objects no longer require a `"type": "sequence"`. That's because these two types of objects are no longer used in a heterogeneous context. (I see this as a big win, as the `"type"` thing always felt inelegant!)

I've updated all example documents to use the new syntax. Use [this link](https://cdn.githubraw.com/w3c/mnx/sequence-simplification/docs/mnx-reference/examples/) to view the examples in context of this branch.

Community members, please have a look/think about these changes, and let me know your feedback! Personally I'm feeling very good about it.

One aspect I'd specifically like feedback on is the new "position" type. I used a simple two-element array because it's nice and compact — but a proper object might be a better approach here.

### DIFF
--- a/docgenerator/data.json
+++ b/docgenerator/data.json
@@ -3479,24 +3479,13 @@
 },
 {
     "model": "spec.jsonobject",
-    "pk": 138,
-    "fields": {
-        "name": "rhythmic position",
-        "slug": "rhythmic-position",
-        "schema": 1,
-        "object_type": 2,
-        "description": ""
-    }
-},
-{
-    "model": "spec.jsonobject",
     "pk": 139,
     "fields": {
         "name": "positioned clef",
         "slug": "positioned-clef",
         "schema": 1,
         "object_type": 1,
-        "description": "<p>Represents a clef as positioned at a specific location within a measure.</p>"
+        "description": "<p>Represents a clef as positioned at a specific location within a measure.</p>\r\n\r\n<p>If the position isn't encoded, the position is assumed to be the start of the measure.</p>\r\n\r\n<p>Note: This object will eventually include a \"position\" key, but we haven't yet designed how that will be encoded.</p>"
     }
 },
 {
@@ -5653,17 +5642,6 @@
         "child": 19,
         "is_required": true,
         "description": ""
-    }
-},
-{
-    "model": "spec.jsonobjectrelationship",
-    "pk": 217,
-    "fields": {
-        "parent": 139,
-        "child_key": "position",
-        "child": 138,
-        "is_required": false,
-        "description": "The clef's position in the measure. If absent, this value is assumed to be [0, 1], the start of the measure."
     }
 },
 {
--- a/docgenerator/data.json
+++ b/docgenerator/data.json
@@ -2258,17 +2258,6 @@
 },
 {
     "model": "spec.jsonobject",
-    "pk": 18,
-    "fields": {
-        "name": "measure contents",
-        "slug": "measure-contents",
-        "schema": 1,
-        "object_type": 2,
-        "description": ""
-    }
-},
-{
-    "model": "spec.jsonobject",
     "pk": 19,
     "fields": {
         "name": "clef",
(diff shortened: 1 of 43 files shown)


williamclocksin:
Yes, I also have a worry about the corner-cases. In general, I like the design principle that elements aren’t constrained unnecessarily. So, in Calliope, clefs and time signatures and barlines etc are treated like any other staff object but just don’t have a time value. It is up to the formatting algorithm to detect when the clef etc is encountered and to position it accordingly taking the timed objects of all the voices into account. Kind regards, William--  (from mobile phone, so forgive brevity)  William ClocksinOn 18 Sep 2023, at 18:26, Christina Noel ***@***.***> wrote:﻿
My only problem with this is the strange edge-case that a clef ends up in the middle of a "zero-time" sequence, like a grace-note run. (See #45 ) also, to make sure they behave properly in the middle of a beam, or when they should only apply to a sequence. (See #96 )
I haven't looked through this carefully to see if it can cover it, since I'm swamped at work right now, but I wanted to make sure these edge cases weren't lost.
Thanks,
Christina

adrianholovaty:
@clnoel — Thanks, that's an excellent point about clef changes within a sequence of grace notes!

Using "appears directly before note-reference-ID" would be a reasonably good solution. The only other option that comes to mind is using a numeric index, like "appears directly before the third event in the sequence," but that's problematic due to nested events (tuplets and grace notes), plus it's more brittle. Are there any other possible solutions for this?

clnoel:
The only other solution that comes to mind is to attach a clef to a sequence, with the understanding that it always occurs at the beginning of the sequence. Any time you need a clef change you stop your sequence and start a new one. This has the side-effect of deciding that a clef is always associated with a sequence, though, which I'm not sure we want. And we'd have to delicately define how sequences of grace notes either stack or become sequential.

--Christina

cecilios:
In this example from Claude Debussy "La Danse de Puck" there is a clef associated to a sequence:
image

Reference taken from Donal Byrd “Gallery of Interesting Music Notation”, http://homes.sice.indiana.edu/donbyrd/InterestingMusicNotation.html

mscuthbert:
I am thinking about how parsing will work with all these elementIds appearing before the element itself is encountered. We may need to have guidelines for implementers.

adrianholovaty:
Thanks for everybody for the comments! I've updated the branch to remove the "position" key, which clearly needs more thought and discussion. The good news is that this only affects inner-bar clef changes; any clef at the start of a measure can be encoded by simply _not_ adding a position.

In the interest of moving this forward, I'm going to merge it and create a new issue for encoding inner-bar clef changes. That will also help simplify my current work in getting the mnxconverter tool up-to-date with our latest spec.

adrianholovaty:
Just a note that I've opened issue #313 to have a more narrowly focused discussion on inner-bar clef changes.

PR #322, opened 2024-02-22
Add event markings, including articulations and single-note tremolos

adrianholovaty:
This pull request adds a `"markings"` object to events. This will hold all event-level markings, such as articulations and tremolos.

Note that it doesn't hold any note-specific markings, such as plops or scoops. Those will likely go into a separate note-specific markings object, to come in future work. There might be an overlap between event-level markings and note-level markings, but we hope to keep it as small as possible, to encourage One And Only One Way to encode this data.

The currently defined markings are:

* accent
* breath
* softAccent
* spiccato
* staccatissimo
* staccato
* stress
* strongAccent
* tenuto
* tremolo
* unstress

The semantics have mostly been copied from MusicXML's [articulations](https://w3c.github.io/musicxml/musicxml-reference/elements/articulations/) and [ornaments](https://w3c.github.io/musicxml/musicxml-reference/elements/ornaments/).

Note: using the term "markings" is very intentionally choosing a term that (1) doesn't have a lot of prior "baggage" and (2) can hold a wide variety of notations, saving users from having to worry about whether a particular notation is an articulation or ornament or whatever. (In some cases this distinction is instrument-specific! So it's best to just sidestep that unnecessary complexity.)

You can view the updated docs (in context of this branch) here:
https://cdn.githubraw.com/w3c/mnx/event-markings/docs/mnx-reference/objects/event-markings/

This also includes two new example documents:

* [Articulations](https://cdn.githubraw.com/w3c/mnx/event-markings/docs/mnx-reference/examples/articulations/)
* [Single-note tremolos](https://cdn.githubraw.com/w3c/mnx/event-markings/docs/mnx-reference/examples/single-note-tremolos/)


mscuthbert:
LGTM!

mscuthbert:
A really cool part about the MNX definition here is that markings are an object with keys rather than a list/array, so there is no possibility of creating an event with two staccatos, or three legatos.

Detached-legato is to be encoded as staccato + tenuto.

PR #347, opened 2024-07-04
Added 'support' key with 'useAccidentalDisplay'

adrianholovaty:
This adds an optional `"support"` object to the top-level `"mnx"` object. The goal of the `"support"` object is to provide information on how to interpret the current document. It's similar to MusicXML's [\<supports\> element](https://w3c.github.io/musicxml/musicxml-reference/elements/supports/).

At the moment, `"support"` includes a single optional key, `"useAccidentalDisplay"`. This tells consuming software whether notes include `"accidentalDisplay"` keys. It answers the question "Does every note with a visible accidental have `"accidentalDisplay"` set?"

If this is set to true, then consuming applications should only draw an accidental on notes that have an `"accidentalDisplay"` value.

If this is set to false, then consuming applications should use their own algorithm to determine whether any given note's accidental should be drawn.

If this value is not provided, it is to be interpreted as false.

The need for this feature arose in my work on the mnxconverter and the MNX viewer. In both cases, I realized there wasn't a clear way to know whether the lack of `"accidentalDisplay"` meant "don't draw the accidental" or "we haven't bothered to encode whether to draw the accidental."

This pull request includes the following changes:

* Added `"support"` and `"useAccidentalDisplay"` to the spec.
* Updated the note object's spec to say `"accidentalDisplay"` should only be used if `"useAccidentalDisplay"` is set to true.
* Updated two example documents to use `"useAccidentalDisplay"` (the two example documents that actually have rendered accidentals: "accidentals" and "organ-layout").

You can view the updated spec via this link:
https://cdn.githubraw.com/w3c/mnx/supports-accidental-display/docs/mnx-reference/objects/support/

I think everything is pretty straightforward here. I did make a few judgment calls:

* What happens when a document doesn't have `"useAccidentalDisplay": true` yet includes `"accidentalDisplay"` on note objects? I decided this should _not_ be allowed, just to avoid any ambiguity. Alternatively we could say "that's allowed, but behavior is undefined," but I'd like to avoid that.
* I called it `"support"` instead of `"supports"`. This way it's a noun instead of a verb.

### DIFF
--- a/docgenerator/data.json
+++ b/docgenerator/data.json
@@ -1799,6 +1799,17 @@
     }
 },
 {
+    "model": "spec.jsonobject",
+    "pk": 162,
+    "fields": {
+        "name": "support",
+        "slug": "support",
+        "schema": 1,
+        "object_type": 1,
+        "description": "<p>This contains information about how to interpret the data within an MNX document.</p>"
+    }
+},
+{
     "model": "spec.jsonobjectrelationship",
     "pk": 1,
     "fields": {
@@ -4042,7 +4053,13 @@
         "child_key": "accidentalDisplay",
         "child": 135,
         "is_required": false,
-        "description": "Information about the displayed accidental for this note. This is required for any note that has a displayed accidental."
+        "description": [
+            "Information about the displayed accidental for this note.",
+            "",
+            "If this MNX document has \"useAccidentalDisplay\" set to true in its top-level \"mnx\" object's \"supports\" object, this \"accidentalDisplay\" key is required for any note that has a displayed accidental.",
+            "",
+            "If \"useAccidentalDisplay\" is missing or set to false, notes must not include \"accidentalDisplay\"."
+        ]
     }
 },
 {
@@ -4390,6 +4407,36 @@
         "child": 105,
         "is_required": false,
         "description": "In a multi-staff part, this specifies which staff the clef is in. If unspecified, this should be interpreted as 1."
+    }
+},
+{
+    "model": "spec.jsonobjectrelationship",
+    "pk": 245,
+    "fields": {
+        "parent": 126,
+        "child_key": "support",
+        "child": 162,
+        "is_required": false,
+        "description": "Information about how to interpret the data in this MNX file, in cases of ambiguity."
+    }
+},
+{
+    "model": "spec.jsonobjectrelationship",
+    "pk": 246,
+    "fields": {
+        "parent": 162,
+        "child_key": "useAccidentalDisplay",
+        "child": 112,
+        "is_required": false,
+        "description": [
+            "This answers the question \"Does every note with a visible accidental have accidentalDisplay set?\"",
+            "",
+            "If this is set to true, then consuming applications should only draw an accidental on notes that have an \"accidentalDisplay\" value.",
+            "",
+            "If this is set to false, then consuming applications should use their own algorithm to determine whether any given note's accidental should be drawn.",
+            "",
+            "If this value is not provided, it is to be interpreted as false."
+        ]
     }
 },
 {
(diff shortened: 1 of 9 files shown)


paul-bayleaf:
I think introducing this implementation defined feature. i.e. "use their own algorithm", is a mistake.  MNX is an interchange format and should specify everything that can be reasonably specified.   Otherwise it damages portability and puts an additional burden on an MNX processor.

Surely all that is required is that if accidentalDisplay is omitted then **define** the default to be don't show the accidental.   If an MNX author fails to use an accidentalDisplay when required then it's simply a bug in their MNX.

adrianholovaty:
@paul-bayleaf Thanks for the thoughts. Much appreciated.

Yes, I suppose it's a question of "where should the burden lie?" — with creating applications or consuming applications?

If we adapt `useAccidentalDisplay` (as in this pull request), then MNX creators aren't forced to calculate accidental display in order to encode a document. Consumers are forced to do so, only for any document that lacks `useAccidentalDisplay=true`.

If we don't adapt `useAccidentalDisplay` (forcing all visual accidentals to always be encoded), then MNX creators are forced to calculate accidental display when encoding. Consumers can rely on that data always being available.

Putting aside the case of "both creator and consumer applications have implemented an accidental-display algorithm," here are some use cases in which either the creator or consumer wouldn't necessarily have such an algorithm:

1. Audio-focused software such as DAWs: they might not care about whether a given note's accidental is graphically displayed, perhaps because they don't even _have_ a notation view. Should they need to implement an accidental-display algorithm purely to export data as MNX?
2. Music-analysis software such as used by computational musicologists: same as 1.
3. Simple sheet music rendering engines: they might not have implemented an accidental-display algorithm. Should they need to implement an accidental-display algorithm to render an MNX document?

This is just a quick list of the top of my head. I'm very interested to hear other people's thoughts on other use cases that would be affected by this.

My personal take is: I believe that having an accidental-display algorithm is a "table stakes" requirement for any software that displays music notation. I also believe there are valid use cases for MNX that operate outside the scope of graphics and shouldn't be burdened with calculating the _graphical decision_ of whether a note has a visual accidental.

In short, I believe the burden is on consuming applications. But, again, I'd like to hear other people's thoughts on this.

One final thought: we could publish some open-source code that populates `accidentalDisplay` for each note in an MNX document, per conventional accidental-display rules. Any software that doesn't have its own algorithm could use this when importing or exporting.

samuelbradshaw:
I don't feel strongly, but I tend to agree with @paul-bayleaf, for this reason: I think you're more likely to have musically-trained experts on the engraving / sheet music creation side than the displaying side, and the burden of following music best practices should be on the side that's closest to the music creation experts.

The displaying side is downstream, and is likely to be thrown together by developers who didn't have any part in the music creation process. I wouldn't trust developers so far from the music creation process to notice engraving issues and/or guess at what the creator intended.

Re: DAWs – a DAW is going to have to do a lot of guessing to be able to export MNX sheet music. For example, MIDI doesn't include the key of the song. The DAW would need to calculate the key, and if it can do that, it can probably calculate accidentals too.

paul-bayleaf:
A further thought: if my suggestion is adopted, that the presence of accidentalDisplay itself indicates that the accidental is to be displayed, then its property "show" becomes redundant and could be removed from the MNX specification.

mscuthbert:
@adrianholovaty -- this sounds great!  I think it would enable a huge number of other applications for MNX, from MIDI/recording systems that don't have a full rendering engine to the kinds of computational musical analysis things that I do.    We could have a fallback position such that a system that can't determine accidental display on its own--when it gets a score created by a similar type of system--should default to showing every accidental.

adrianholovaty:
I've just merged this, as the three co-chairs are in unanimous agreement about the idea.

Issue #354, opened 2024-09-11
Basic lyrics support


adrianholovaty:
Just to get a discussion going, here's a rough first draft of what lyrics support in MNX might look like.

Obviously there's a lot to lyrics — formatting, language encoding, extender (melisma) lines, non-word sounds such as MusicXML's `<laughing>` and `<humming>`. I'd like to design this in a way that gets us a solid foundation that we can elegantly extend later, rather than getting forever bogged down in obscure features.

For prior art, see MusicXML's approach to lyrics:
https://w3c.github.io/musicxml/musicxml-reference/elements/lyric/

My approach here is to sketch out a simple example, then provide some thoughts on how it could be extended to support more obscure aspects of lyrics.

## Proposed simple example

<img width="418" alt="Screenshot 2024-09-11 at 8 36 48 PM" src="https://github.com/user-attachments/assets/db85e3c2-5e23-4dda-b98c-030da2a398b9">

```json
[
  {
    "type": "event",
    "duration": {"base": "quarter"},
    "lyrics": {
      "lines": [
        {"text": "Are"}
      ]
    },
    "notes": [
       {"pitch": {"octave": 5, "step": "C"}}
    ]
  },
  {
    "type": "event",
    "duration": {"base": "quarter"},
    "lyrics": {
      "lines": [
        {"text": "you"}
      ]
    },
    "notes": [
       {"pitch": {"octave": 5, "step": "D"}}
    ]
  },
  {
    "type": "event",
    "duration": {"base": "quarter"},
    "lyrics": {
      "lines": [
        {"text": "sleep", "hyphen": "start"}
      ]
    },
    "notes": [
       {"pitch": {"octave": 5, "step": "E"}}
    ]
  },
  {
    "type": "event",
    "duration": {"base": "quarter"},
    "lyrics": {
      "lines": [
        {"text": "ing?", "hyphen": "end"}
      ]
    },
    "notes": [
       {"pitch": {"octave": 5, "step": "C"}}
    ]
  }
]
```

A pretty basic starting point. Lyrics would live on events. Each event would have an optional `"lyrics"` key, which would be a lyrics object. Its `"lines"` key would be an array of lyric syllable objects. Each lyric syllable is required to have a `"text"`. (Eventually I could envision loosening this: each lyric syllable would be required to have a `"text"` or something like `"laugh": true`.)

## Hyphens

[...]

Issue #10, opened 2017-04-26
Handle interpolated text blocks


joeberkovitz:
An important feature of CWMN scores is the concept of an interpolated text block, such as lyric verses or instructions, that are presented separately from systems of parts. In fact, these text blocks can be thought of as "textual systems" that are flowed into the paginated music at the appropriate places. These text blocks are typically anchored semantically to some point in the score, so that if the score is reflowed, the text block still shows up in a meaningful place. 

MusicXML lacks a mechanism for handling these, other than explicit insertions of text on specific pages at specific X/Y positions, which doesn't capture the semantic content obviously. MNX should support this important feature.

mdgood:
This is similar to [MusicXML issue 40](https://github.com/w3c/musicxml/issues/40) which we agreed to defer to MNX. I would like to see this issue generalized to include graphics similar to how the MusicXML issue was worded.

MusicXML can also handle these systems by attachment to the "nearest measure" rather than a specific place on the page. However that still does not address the semantics that these text and graphics are reflowable counterparts to musical systems.

PR #355, opened 2024-09-26
Added spec for lyrics and lyric lines, with two new example docs

adrianholovaty:
This implements basic support for lyrics:

* Events now have a ["lyrics" key](https://cdn.githubraw.com/w3c/mnx/lyrics/docs/mnx-reference/objects/lyrics/), which is a lyrics object.
* The lyrics object has a "lines" key, which is a list of [lyric lines](https://cdn.githubraw.com/w3c/mnx/lyrics/docs/mnx-reference/objects/lyric-line/).
* A lyric line has a required "text" key, plus an optional "continues" boolean.

This includes two new example docs:

* [Lyrics (basic)](https://cdn.githubraw.com/w3c/mnx/lyrics/docs/mnx-reference/examples/lyrics-basic/)
* [Lyrics (multi-line)](https://cdn.githubraw.com/w3c/mnx/lyrics/docs/mnx-reference/examples/lyrics-multi-line/)

Note that this changes the JSON Schema to define a custom "string" type, which is not intentional. I need to tweak the docs system not to do that. I'll take care of that before merging this pull request, but in the meantime I'd like to get feedback on this basic lyrics support.


williamclocksin:
There should be two type of continuations: hyphen (which you have specified) and melisma. On 26 Sep 2024, at 19:01, Adrian Holovaty ***@***.***> wrote:﻿This implements basic support for lyrics:

Events now have a "lyrics" key, which is a lyrics object.
The lyrics object has a "lines" key, which is a list of lyric lines.
A lyric line has a required "text" key, plus an optional "continues" boolean.

This includes two new example docs:

Lyrics (basic)
Lyrics (multi-line)

Note that this changes the JSON Schema to define a custom "string" type, which is not intentional. I need to tweak the docs system not to do that. I'll take care of that before merging this pull request, but in the meantime I'd like to get feedback on this basic lyrics support.



You can view, comment on, or merge this pull request online at:
  https://github.com/w3c/mnx/pull/355

Commit Summary

  4045f19  Added spec for lyrics and lyric lines, with two new example docs


File Changes  (36 files)

  
    M
    docs/mnx-reference/examples/index.html
    (18)
  
  
    A
    docs/mnx-reference/examples/lyrics-basic/index.html
    (198)
  
  
    A
    docs/mnx-reference/examples/lyrics-multi-line/index.html
    (211)
  
  
    M
    docs/mnx-reference/objects/boolean/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/clef-sign/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/clef/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/event/index.html
    (17)
  
  
    M
    docs/mnx-reference/objects/global/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/index.html
    (4)
  
  
    A
    docs/mnx-reference/objects/lyric-line/index.html
    (117)
  
  
    A
    docs/mnx-reference/objects/lyrics/index.html
    (109)
  
  
    M
    docs/mnx-reference/objects/measure-global/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/mnx/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/note-value-base/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/note-value/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/note/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/octave/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/part-measure/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/part/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/pitch/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/positioned-clef/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/positive-integer/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/root/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/sequence/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/staff-position/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/step/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/string/index.html
    (7)
  
  
    M
    docs/mnx-reference/objects/time-signature-unit/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/time/index.html
    (2)
  
  
    M
    docs/mnx-reference/objects/version-number/index.html
    (2)
  
  
    M
    docs/mnx-schema.json
    (29)
  
  
    A
    docs/static/examples/lyrics_basic.png
    (0)
  
  
    A
    docs/static/examples/lyrics_multiline.png
    (0)
  
  
    M
    doctools/data.json
    (936)
  
  
    A
    doctools/media/examples/lyrics_basic.png
    (0)
  
  
    A
    doctools/media/examples/lyrics_multiline.png
    (0)
  


Patch Links:

  https://github.com/w3c/mnx/pull/355.patch
  https://github.com/w3c/mnx/pull/355.diff

mscuthbert:
About:

> Rename "continues": true to "hyphen": true — to avoid possible confusion with extender lines. (Those also "continue" a note, of course!)

One reason for explicitly noting that a *hyphen* will result rather than another semantic encoding like `continues` or `mid-word = true` is that multi-syllable/character words in many Asian languages should not use hyphens.

paul-bayleaf:
If I've understood the revised proposal where each event's lyrics is an array of objects each with a user defined key then, without a specified ordering of the keys, in order to determine on which line a lyric should be placed the MNX processor has to first collect **all** the user defined lyric keys for the **entire** piece and then sort them.   Only then can the lyrics' lines be established.    An array, as per your original proposal, immediately gives the line position, which I think is a lot simpler to deal with.  What is so awkward about empty array elements?    Alternatively, if objects are to be used then rather that having a user defined key require that the key is the lyric's line number.

samuelbradshaw:
> About:
> 
> > Rename "continues": true to "hyphen": true — to avoid possible confusion with extender lines. (Those also "continue" a note, of course!)
> 
> One reason for explicitly noting that a _hyphen_ will result rather than another semantic encoding like `continues` or `mid-word = true` is that multi-syllable/character words in many Asian languages should not use hyphens.

To make sure I'm understanding correctly – will a hyphen always be drawn on the page if `"hyphen": true` and never be drawn on the page if `"hyphen": false`? Or is `"hyphen": true` a semantic property that says the word continues, but whether or not a hyphen symbol is drawn depends on the language?

If the latter, I'm worried that this change will make it harder to extract lyrics from sheet music. For example, in Korean sheet music like this:
https://assets.churchofjesuschrist.org/f3/14/f3141a6cd94456a64017fd99fce71e7e384b1a98/high_on_the_mountain_top.pdf
Ideally I'd be able to extract lyrics like this from the MNX sheet music:
1. 높은 산 언덕 위
깃발이 날린다
온 땅 위에 덮이니
만 백성 보네
꿀벌의 땅 평화의 곳
시온의 언덕 위에 서있네

But if we're moving away from a semantic indicator that tells me if the word ends or not, is there any way I can know where to put spaces?

adrianholovaty:
@paul-bayleaf Thanks for the thoughts. I neglected to mention another reason to avoid using arrays, which is the possibility of lyrics above the staff. See the "Lyrics above the staff vs. below" section in issue #354. Those could be encoded as a separate `"linesAbove"` array, but I think it would be more elegant to store all the lyrics lines together, as the "above vs. below" decision is a bit presentational in nature.

There are also some concrete benefits in giving each line an addressable name, so that we can attach document-level metadata to them. Specific ideas for metadata about a lyric line:

* Its position (above or below the staff)
* Its order (if deviating from default, which would be alphabetical by line ID string)
* Its language (e.g., French)
* Its visual formatting (e.g., italic)

We could also have a concept of "lyric spans," where the document would define a given range of music ("bars 1-16") with some metadata about how the lyric lines should behave in the range.

Apologies, I should have written up those ideas, as I realize we talked about them in our co-chairs meeting but I didn't communicate them here.

Responding to your point about needing to calculate the number of lyrics lines in order to display the music: you'd only need to do this for a given system at a time (not the entire score). If some parts of a score have four lyric lines, but the "chorus" only contains a single line, and the chorus is on its own system, then that system would only need to allocate space for the single "chorus" line. Let me know if I'm misunderstanding the question.

samuelbradshaw:
I have some ideas of how to make lyrics easier to extract – but I can open that as a separate issue. I don't have any objections on my end to moving forward with this pull request for basic lyrics support. Thanks for all your work on this!

adrianholovaty:
@samuelbradshaw OK, great to hear it. We definitely appreciate your expertise in this area!

Regarding your `"hyphen"` question, let me get back to you with a proper answer. @mscuthbert had the idea to rename it and I'll be chatting with him later today.

mscuthbert:
Ah. I had not realized there were languages that did not use hyphens in musical lyrics but also used spaces when extracted. The chairs had discussed Chinese and Japanese (and Hebrew) but not Korean. 

Expertise in Arabic lyric setting would also be a help if anyone has it.

samuelbradshaw:
> Responding to your point about needing to calculate the number of lyrics lines in order to display the music: you'd only need to do this for a given system at a time (not the entire score). If some parts of a score have four lyric lines, but the "chorus" only contains a single line, and the chorus is on its own system, then that system would only need to allocate space for the single "chorus" line. Let me know if I'm misunderstanding the question.

Note that the chorus doesn't always start in a new system. For example:
https://assets.churchofjesuschrist.org/b4/10/b4109e8e18972270c458ac9134bfd715ad4f0c96/battle_hymn_of_the_republic.pdf

adrianholovaty:
Just had the biweekly co-chairs chat and we decided to change this pull request a bit more:

* Instead of `"hyphen"`, use `"type"`. This would be `"whole"` (default), `"start"`, `"middle"` or `"end"`. The previous idea of `"hyphen": true` would now be encoded as either `"start"` or `"middle"`, depending on context. This new approach means there's more data to encode, and it means there's possibility for ambiguity/bad data, but it enables a few more use cases, such as helping a graphics engine decide how close a given lyric can be rendered to another one. We would explicitly document what MNX consumers would be expected to do in each case of "bad" data (@mscuthbert created a quick rubric for this).
* Rename "lyric line" to "event lyric line" and use the term "lyric line" to refer to the _entire_ line across the full score. We didn't previously have a term for this, and it's generally useful to have a term.

mscuthbert:
Adrian will report more, but after discussing these corner cases, it looks like a 4-way enumeration like MusicXML's syllabic will be retained.  ( https://www.w3.org/2021/06/musicxml40/musicxml-reference/data-types/syllabic/ )  -- no names determined at present.  Given this, and given that mismatched `<syllabic>` etc. is one of the most common error conditions in music transfer, I'd like to propose a translation table for impossible combinations that consuming applications should use.  Assuming the names are "start" (S) "mid" (M) "end" (E) "whole" (W)  (musicxml "begin" "middle" "end" "single"), and assuming one-distance look-behind (store last value) with only the second/current value changeable:

```
S S -> S M     M S -> M M     E S -> ok      W S -> ok
S M -> ok      M M -> ok      E M -> E S     W M -> W S
S E -> ok      M E -> ok      E E -> E W     W E -> W W
S W -> S E     M W -> M E     E W -> ok      W W -> ok
```

A consequence of this table is that applications that can only encode a boolean "hyphen-follows" value can encode true/false as S + W and then (I believe) can walk through the data a second time to convert to S,M,E,W values that are consistent with that data.

samuelbradshaw:
> Expertise in Arabic lyric setting would also be a help if anyone has it.

Re: Arabic, I've seen two different styles, but I'm not sure which is preferred by native speakers.

1. Lyrics inline with the music, with the musical notation right to left, as in this book:
https://www.churchofjesuschrist.org/bc/content/shared/content/arabic/pdf/language-materials/33438_ara.pdf#page=7
2. Lyrics below the music:
https://assets.churchofjesuschrist.org/23/0b/230bffd9d2583a84bcc420f3ab5308dd7abcbb0e/count_your_blessings_ara.pdf

Neither has hyphens. I'm not sure if any modern notation software natively supports the right-to-left music notation style.

Languages like Thai and Khmer/Cambodian are kind of the opposite of Korean. Whereas Korean has spaces in lyrics, but no hyphens in the sheet music, Thai and Khmer have hyphens in sheet music, but no spaces in the lyrics. Here's a Khmer example:
https://assets.churchofjesuschrist.org/59/a1/59a17ff8953f52801ad77aca927ec5c6d92cf832/the_morning_breaks.pdf
But the verse written out doesn't have spaces:
ថ្ងៃ​រះ​ឡើង​ហើយ​ស្បៃ​ង​ងឹត​ឃ្លាត
មើល​ទង់​ស៊ី​យ៉ូន​ត្រ​ដាង​លា​លាត!
ព្រ​លឹម​ស្រាង​ៗ​នៃ​ថ្ងៃ​រះ​ថ្មី
ព្រ​លឹម​ស្រាង​ៗ​នៃ​ថ្ងៃ​រះ​ថ្មី
ព្រះ​ចេ​ស្តា​ទ្រង់​ចាំង​ពេញ​ផែន​ដី។

mscuthbert:
Thank you Samuel! Wow. What an amazing set of resources. There are cautions in using one publisher for so many different language resources (whether a pan-editorial condition overrides local practice) but I think everyone in the field has very high respect for LDS hymnal editing practices. 

And thankfully lyrics are one of the few places where Mongolian does not pose especially hard or unique problems (as the world's sole(?) remaining strictly vertical only language -- unlike Japanese/Chinese etc which can be written either way.)

samuelbradshaw:
> There are cautions in using one publisher for so many different language resources (whether a pan-editorial condition overrides local practice) but I think everyone in the field has very high respect for LDS hymnal editing practices.
Thank you – agreed, I'd love to see examples in other languages from a wide range of publishers.

paul-bayleaf:
"Its order (if deviating from default, which would be alphabetical by line ID string)"

Alphabetical ordering of Unicode is not a trivial matter!   It is locale dependent. E.g.
Swedish: 	z < ö
German: 	ö < z
For MNX surely the ordering must be locale independent, otherwise we lose portability.    If we go down this path we will need to be very precise as to what "alphabetical" means. 

Unicode® Technical Standard #10 Unicode Collation Algorithm
https://www.unicode.org/reports/tr10/

williamclocksin:
One question:  It appears that the provision for lyrics assumes that all the lyric syllables on a given line are in the same font.  Is there ever any call for putting one or more of the syllables on the line in a font (i.e. name, weight, posture) other than the one specified in the lyric line?

W.

adrianholovaty:
@williamclocksin We don't have a way to specify fonts for lyrics yet, but, yes, that's definitely something we should aim to support.

PR #488, opened 2026-01-13
Update examples to latest json

rpatters1:
This PR updates 2 example json files that now fail to validate against the schema. (This is because part.measures is now required.)

FWIW: my validator identifies a number of other issues with these files as well as the organ-layout file. References to non-existent measures, missing first clefs. That kind of thing. But this PR includes only those changes needed to get the files to validate against the schema.

### DIFF
--- a/docs/static/examples/json/orchestral-layout.json
+++ b/docs/static/examples/json/orchestral-layout.json
@@ -2,19 +2,19 @@
 "mnx": {"version": 1},
   "global": {"measures": []},
   "parts": [
-    {"id": "Klar1"},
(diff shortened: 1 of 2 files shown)


adrianholovaty:
Thanks, I've integrated these changes into the docs system and committed in 755518757.

Those two examples have always been a bit odd, since they're so "snippet-y." We should probably bite the bullet and encode more of the music to prevent confusion. My original intent was to make the example deliberately simple so as not to be distracting, but that's less of a concern now that our docs system can highlight specific parts of the markup.

PR #489, opened 2026-01-14
Create example for `tie.targetType`.

rpatters1:
The example

- validates against the schema
- validates against my semantic validator
- produces the expected result in my importer (shown)

Screenshot 2026-01-13 at 7 24 48 PM

### DIFF
--- a/docs/static/examples/json/tie-target-types.json
+++ b/docs/static/examples/json/tie-target-types.json
@@ -48,7 +48,6 @@
                                         "base": "quarter"
                                     },
                                     "id": "ev21",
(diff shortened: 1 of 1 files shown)


adrianholovaty:
Thanks very much! Great to get that "lv" example in there too.

I've hooked this JSON into the docs system to make everything crosslinked, and I added some `"_x.mnxdocs"` objects so that the relevant bits of the document are highlighted in blue in context of the HTML view. Here's the new page:

https://w3c.github.io/mnx/docs/mnx-reference/examples/tie-targets/

PR #500, opened 2026-02-07
Keyfifth flip examples

rpatters1:
This PR proposes two new examples to clarify edge cases around `keyFifthsFlipAt` in `part-transposition`. I would have preferred to combine them in a single example, but my tools (MuseScore and Finale) each can only do part of the example.

- `keyfifths-flip.json` illustrates flipping around the edge case values `5`, `7`, and `8`. The result is this:

Screenshot 2026-02-07 at 7 48 02 AM

- `keyfifths-flip-2.json` illustrates what happens if `keyFifthsFlipAt` is omitted:

Screenshot 2026-02-07 at 8 05 42 AM

One reason I am submitting these examples is that I want to be certain that I have interpreted the text of the specification correctly.

### DIFF
--- a/docs/static/examples/json/keyfifths-flip-2.json
+++ b/docs/static/examples/json/keyfifths-flip-2.json
@@ -1,0 +1,104 @@
+{
+    "mnx": {
+        "version": 1
+    },
+    "global": {
+        "measures": [
+            {
+                "barline": {
+                    "type": "double"
+                },
+                "key": {
+                    "fifths": 4
+                },
+                "time": {
+                    "count": 4,
+                    "unit": 4
+                }
+            },
+            {
+                "key": {
+                    "fifths": 6
+                }
+            }
+        ]
+    },
+    "parts": [
+        {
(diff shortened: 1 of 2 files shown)


rpatters1:
As a total aside in the weeds, here is an actual use case for the Finale behavior that is more of a workaround. Finale has an artificial limit of +/-7 alterations. When the microtonalists set up higher EDOs, e.g., 96-EDO (with 8 divisions per half step), they can't reach a full chromatic half step with the available alterations. The workaround is to change the key so that the needed sharp or flat note is in the key and turn on the "hide key sig and show accis" option. That displays them correctly. Not something we need to worry about here at MNX, though.

rpatters1:
I found an example in the wild of a -8 key signature. The 2nd movement of the Ewald Brass Quintet No. 4 is in Fb Major, which means the non-transposing instruments have 8 flats. Here is the opening of the tuba part:

Screenshot 2026-06-23 at 8 45 51 AM

